PR #2, opened 2016-05-09
Update composer.json

server-may-cry:
php namespaces introduced in [PHP 5.3 version](http://php.net/manual/en/language.namespaces.basics.php)


dcramer:
Thanks!

Issue #7, opened 2016-05-17
Set Current User


bassrock:
In the old monolog raven way, there was a way to create a custom processor that allows you to set the current logged in user and any associated data keys.

How could this be accomplished now?

Jean85:
This should be a very nice feature!
Right now the user is identified by its session, but it may vary if you wipe the sessions or if they expire.

I would suggest to do so automatically in this bundle with:
 ```
$this->get('security.token_storage')->getToken()->getUser();
```
It should just be an overridabile behaviour.

dcramer:
I'd love if we could find a way to expose the Sentry client.

e.g. does ``$this->get('sentry.client')`` work?

That'd make it easy to do everything the client allows, but I'm not familiar enough with Symfony to know if there's a good pattern for this.

Jean85:
I'm trying to work on it. Right now I'm just adding a functionality that sets the username as the user ID in the raven client, but I'm not familiar enough with Raven... is a string ok as an ID?

Anyhow, `$this->get('sentry.client')` works, so yes, it's exposed right now. I'll open a PR to discuss this ASAP.

Jean85:
Ok, my PR is merged. It uses the username as a user ID inside Sentry, but I think that more could be achieved.

The interesting approach should be an extensible one, where the dev can write a class that somehow provides additional information (user email at the very least!). I currently don't know how to achieve this.. Any suggestions?

dcramer:
@Jean85 I think we should empower people to just pull in the client and write their own middleware if defaults dont work for them. The solution might simply be to document how to do this.

Jean85:
Yeah that could be a way. Anyhow, we should suggest a way that doesn't collide with the standard approach. Now we are using a listener, and a user should use an other listener for his custom implementation; maybe we can force the listener priority?

swquinn:
@dcramer and @Jean85 see my above proposed PR (#47) I think through using the event dispatcher and trigger certain events, e.g. `SET_USER_CONTEXT` and `PRE_CAPTURE` you can allow projects to simply tap into and set their own values using a subscriber/listener pattern to override the defaults set.

Of course, these projects already have the option to completely customize the client and exception listener classes if they want to, but that is a little more extreme.

My PR also includes an update to the `README.md` to illustrate how the above can work.

Jean85:
#47 is now merged, and it can be widely used to do all kinds of customization on the event information, including the user!

florentdestremau:
[...]

PR #9, opened 2016-05-18
Setting the user through a new listener

Jean85:
As I was saying in #7 (which this addresses) I'm working on a standard way to set the user during the normal request cycle. It's not complete, but I think that's a start.

### DIFF
--- a/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
+++ b/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
@@ -2,33 +2,106 @@
 
 namespace Sentry\SentryBundle\EventListener;
 
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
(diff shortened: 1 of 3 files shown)


Jean85:
As a side note, my approach was inspired by the Doctrine Extensions: https://github.com/Atlantic18/DoctrineExtensions
In particular: https://github.com/stof/StofDoctrineExtensionsBundle/blob/master/EventListener/BlameListener.php#L54

src/Sentry/SentryBundle/EventListener/ExceptionListener.php
```diff
@@ -2,33 +2,106 @@
...
             return;
         }
 
         $this->client->captureException($exception);
     }
+
+    /**
+     * @param UserInterface | object | string $user
+     */
+    private function setUserValue($user)
```
dcramer: is it guaranteed in Symfony to actually have a confirmed user interface? Otherwise maybe we should try/catch already this functionality

dcramer:
Ignore my comment :)

Issue #11, opened 2016-06-02
[feature] prevent store some public exceptions


server-may-cry:
Would be nice to have way set wich exception classes not store [here](https://github.com/getsentry/sentry-symfony/blob/0.3.0/src/Sentry/SentryBundle/EventListener/ExceptionListener.php#L84..L87)
HttpExceptionInterface is not enought

PR #12, opened 2016-06-04
Now you can configure a list of exceptions that would not be sent to the sentry

server-may-cry:
#11


dcramer:
This is great, thanks!

PR #15, opened 2016-06-16
Symfony security service should be optional

fain182:
I am using [symfony as a microframework](http://symfony.com/blog/new-in-symfony-2-8-symfony-as-a-microframework) and I don't use the security component.

I don't think it should be a mandatory dependency.


dcramer:
:+1: thanks!

PR #16, opened 2016-06-27
Added and registered a listener method for console exceptions

tbmatuka:
We were missing this type of exception logging, so here's a tiny commit that adds it.


dcramer:
For my own education, console exceptions are just from shell scripts?

tbmatuka:
Yup. Only console scripts should be using them.

dcramer:
Thanks!

PR #21, opened 2016-09-01
Add option and API to disable Sentry reporting on demand

Addvilz:
This PR suggests adding an option and API to disable Sentry reporting - during tests, in testing environment's etc.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-symfony/21)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,10 @@ sentry:
 
 The following can be configured via ``app/config/config.yml``:
 
+### enabled
+
+Enable or disable reporting to Sentry. Defaults to true.
+
 ### app_path
 
 The base path to your application. Used to trim prefixes and mark frames as part of your application.
--- a/src/Sentry/SentryBundle/DependencyInjection/Configuration.php
+++ b/src/Sentry/SentryBundle/DependencyInjection/Configuration.php
@@ -22,6 +22,9 @@ class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
+                ->booleanNode('enabled')
+                    ->defaultTrue()
+                ->end()
                 ->scalarNode('app_path')
                     ->defaultValue('%kernel.root_dir%')
                 ->end()
--- a/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
+++ b/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
@@ -29,19 +29,26 @@ class ExceptionListener
     /** @var  string[] */
     private $skipCapture;
 
+    /** @var bool */
+    private $reportingIsEnabled;
+
     /**
      * ExceptionListener constructor.
      * @param TokenStorageInterface $tokenStorage
      * @param AuthorizationCheckerInterface $authorizationChecker
      * @param \Raven_Client $client
      * @param array $skipCapture
+     * @param $reportingIsEnabled
      */
     public function __construct(
         TokenStorageInterface $tokenStorage = null,
         AuthorizationCheckerInterface $authorizationChecker = null,
         \Raven_Client $client = null,
-        array $skipCapture
+        array $skipCapture,
+        $reportingIsEnabled
     ) {
+        $this->reportingIsEnabled = $reportingIsEnabled;
+
         if (!$client) {
             $client = new SentrySymfonyClient();
         }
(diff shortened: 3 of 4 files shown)


dcramer:
Would this have the same effect as simply setting the DSN to an empty value?

Addvilz:
Likely, but I was not able to find any documentation on what effect that has on the client exactly. This has an added value of being able to disable/enable reporting temporarily - for example, for integration tests, to verify that sentry is being invoked at all times it is meant to be invoked. Plus this is isolated in the bundle and does not rely on "external" API.

Personally, I think having explicit API and flag is way better than mangling with DSN. :) What do you think?

dcramer:
In every Sentry SDK setting a falsely value for he DSN disables reporting (it's a documented standard in our spec). Because that's all this does I think we shouldn't add a duplicate path for that. There is however an argument to be made on having a flag which completely avoids even creating the integration.

Addvilz:
You mean to not create the service in container? How do you feel would be better to approach this?

server-may-cry:
also you can simple use this
```yaml
sentry:
    skip_capture:
        - "Exception"
```

Addvilz:
Ok, closing this then.

ryabenko-pro:
@dcramer, with all the respect, adding this option makes it look much better.
Disabling plugin and removing DSN options are two different actions. This is not explicit way and not correct, since no one will remove the string but comment it, which makes config file messy. This does not affect standards in your spec because symfony bundles and your SDKs are different things. 

My car has neutral gear and I can use it when I want engine on, but car not moving. Please consider reopen this PR. 

Thank you.

dcramer:
@ryabenko-pro software has nothing to do with your cars gears, and we have a standard in place for "how to disable sending errors with Sentry" that is already followed. There's little to no value in adding a second mechanism to disable Sentry just because of personal preference.

ryabenko-pro:
Thank you @dcramer, very clear and loud. 

Software and car building have a lot in common - both are usually done by engineers, but this is rather personal opinion. Personal opinion of your potential customer, btw.

ryabenko-pro:
Here is real case - using .env provided by symfony sentry is crashing dev env, because I can't pass null there. I'm getting exception:

    InvalidArgumentException in Client.php line 429:
    Unsupported Sentry DSN scheme: <not set>

I've been trying following options:
    
    # .env file
    SENTRY_DSN=null
    SENTRY_DSN=NULL
    SENTRY_DSN=~
    SENTRY_DSN=""
    SENTRY_DSN=

result in same exception. I have to go and change `parameters.yml` content which is not really good. 
Please consider providing optional, not BC breaking `disable` option. 

Thank you. 

P.S.: If you have recipe how to make it working with very standard S3 DotEnv component, I'll leave this topic.

dcramer:
My expectation is that ``SENTRY_DSN=`` would work. It's possible somethings wrong and thats not true. The others (barring the empty string), are not valid, as you're simply passing real values in which would be invalid DSNs.

Jean85:
@ryabenko-pro I think that you are misusing the Symfony Dotenv component.
Are you sure that your `parameters.yml` isn't overwriting what you would like to do with your .env file?

ryabenko-pro:
@dcramer, @Jean85 sorry for disturbing. Can't reproduce the issue. 

I've used `sentry.dsn: '%env(SENTRY_DSN)%'` in `parameters.yml`, but had to use just `sentry.dsn: ~`. Now empty string and `""` are working. 

Anyways, thank you for your time, please disregard my previous message.

ryabenko-pro:
So the issue were found. Bitbucket pipeline uses image with weird feature: empty value in .env file being parsed as string with one space " " (see screenshot). Sentry thinks that a space is not empty DSN, start parsing it and throws exception "Unsupported Sentry DSN scheme: <not set>". Locally and on qa/prod servers any amount of spaces as a value results in empty string, so issue only appears on Bitbucket pipelines. If I put "http://sentry.ux/sucks" as DSN value - everything works perfectly, but if someone register that domain name testing errors with context will leak, so I decide to write to Bitbucket support rather than leaving valid domain in DNS. 

Thanks Sentry team for interesting investigation time, I know how important specs are, no matter if with them commercial project provides user experience lower than some open source projects. 

tursa___bk-main___add-on___pipelines_ _bitbucket

Jean85:
Happy to see that you resolved it!
Maybe we can throw a `trim()` to that value to solve this (and similar) issues?

ryabenko-pro:
@Jean85 

unfortunately I did not yet resolved it. `trim` would do the trick I'm sure.

Jean85:
@ryabenko-pro just released the fix: https://github.com/getsentry/sentry-symfony/releases/tag/0.8.5

ryabenko-pro:
@Jean85 

Thank you very much for fix, despite the tone of my comment. I edited my comment to remove bad behaviour example.

PR #22, opened 2016-09-01
Exclude examples from distribution

Addvilz:
This PR suggests exclusion of examples directory from git distribution package (so, no example dir in vendors, except if installed with prefer source)

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-symfony/22)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+examples/ export-ignore


dcramer:
Thanks!

Issue #31, opened 2016-10-24
Exception are grouped the wrong way due to wrong stacktrace


Jean85:
We are using 0.5 and sentry/sentry 1.5.0 in production, and we are currently seeing a strange behavior in Sentry: it's grouping together exceptions that came from very different places, just because they are the same classes (`\LogicException`).

I thing the culprit is the stacktrace.

Due to how Symfony is build, there is a `$response = call_user_func_array($controller, $arguments);` in every handled request. If do not ask to see the full stacktrace, Sentry shows only it up to that line, hiding al the code that belong to my project; I think this leads to the wrong grouping.

Screenshot
![sentry-screenshot](https://cloud.githubusercontent.com/assets/6729988/19637818/288fbf62-99d1-11e6-9602-33ef214cd4a2.png)

Jean85:
Is this maybe a Sentry issue? It seem to ignore anything above the anonymous function...
Paste of the raw stacktrace (with redacted text):

```
LogicException: [redacted]
#8 src/Facile/[redacted].php(1312): [redacted]
#7 src/Facile/[redacted].php(885): [redacted]
#6 src/Facile/[redacted].php(304): [redacted]
#5 [Anonymous function](0): call_user_func_array
#4 app/bootstrap.php.cache(3238): handleRaw
#3 app/bootstrap.php.cache(3197): handle
#2 app/bootstrap.php.cache(3351): handle
#1 app/bootstrap.php.cache(2536): handle
#0 web/app.php(19): null
```

dcramer:
Whats happening here is its not including 'src' in the app paths. Is that a common setup? If so we can add it to the defaults?  Additionally, as I dont think its configurable yet, we could make it so you can specify your own app paths.

dcramer:
Actually, I'm looking at our sample project, and I'm not sure why app path is even set to this.

dcramer:
So this is mostly good, but in the future we should probably come up with a better solution for anonymous functions:

![screenshot 2016-10-24 11 11 07](https://cloud.githubusercontent.com/assets/23610/19657851/9b67578a-99da-11e6-9ba3-26b1074361ac.png)

Jean85:
Yeah, that is a pretty common setup, it's how Symfony is normally installed as a full framework. Thanks for the quick fix! 

[...]

PR #32, opened 2016-10-24
Improve app path handling

dcramer:
Utilize new exclusion list to include top level directory as app path, and exclude vendor subpath.

Fixes GH-31


dcramer:
So this is mostly good, but in the future we should probably come up with a better solution for anonymous functions:

screenshot 2016-10-24 11 11 07

PR #34, opened 2016-11-03
Add error types option

d3f3kt:
According to my pull request: getsentry/sentry-php#369
With this error types options its possible to configure which error types should be
reported.


src/Sentry/SentryBundle/Resources/config/services.yml
```diff
@@ -1,7 +1,7 @@
 services:
     sentry.client:
         class: '%sentry.client%'
-        arguments: ['%sentry.dsn%']
+        arguments: ['%sentry.dsn%', {'errro_types': '%sentry.error_types%'}]
```
Addvilz: Typo: error_types?

test/Sentry/SentryBundle/Test/ErrorTypesParserTest.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
Addvilz: The test cases seems lacking. For configuration extension, `SentrySymfonyClient`, no container test...

src/Sentry/SentryBundle/Resources/config/services.yml
Addvilz: Line 4 has `errro_types`, should have `error_types`. Nothing to do with that change.

README.md
```diff
@@ -112,3 +112,12 @@ sentry:
...
         - "Symfony\\Component\\HttpKernel\\Exception\\HttpExceptionInterface"
 ```
+
+### error types
+
+Define which error types should be reported.
+
+```yaml
+sentry:
+    error_types: E_ALL & ~E_DEPRECATED & ~E_NOTICE
```
Addvilz: Would it not be simpler to just pass array's RavenClient expect instead of transforming strings to constants, back to Sentry levels again?

d3f3kt: Oh, you're right - 'errro_types' was a typo. It's corrected now.

Of course it would be easier, but this is the established syntax to define error levels in php.

Addvilz: It is a way (legacy one, I might add) on how you define error reporting levels for PHP interpreter - nothing more. This configuration however is meant for Sentry, and Sentry has it's own error levels defined already - so it only makes sense, and is in a way expected behavior to follow Sentry convention. 

As far as I know, otherwise, it's inconsistent - Symfony client uses one set of levels, Sentry uses another. What gives?

d3f3kt: I don't really know which Sentry error levels do you mean. The Raven_ErrorHandler exactly needs the error levels how i defined it in the configuration. 
```php
    /**
     * Register a handler which will intercept standard PHP errors and report them to the
     * associated Sentry client.
     *
     * @param bool $call_existing Call any existing errors handlers after processing
     *                            this instance.
     * @return array
     */
    public function registerErrorHandler($call_existing = true, $error_types = null)
    {
        if ($error_types !== null) {
            $this->error_types = $error_types;
        }
        $this->old_error_handler = set_error_handler(array($this, 'handleError'), E_ALL);
        $this->call_existing_error_handler = $call_existing;
        return $this;
    }
```

Addvilz: Yep, terribly sorry, I was under impression RavenClient uses "error", "warning" kind of strings. 

Apart from that - can we avoid at least part of the conversions? For example, instead of having the php.ini-like single variable string `E_ALL && ~E_WARNING`, perhaps we can have array of levels we want and then just get the values via constant? Without all the preg_replace etc.

d3f3kt: Yep of course thats possible, but the configuration would be messed up. Because you cant simple negate Error Types e.g. ~E_WARNING. Instead you have to define every single error type you want to get reported. 
Thats why I thought it's a nice feature to parse and evallute this native way of defining php error levels.

Addvilz: Of course. However, consider that this code is used in projects of all sizes, and this code is executed on construct - which is both good (executed just once) and bad thing (executed always). It's not a big issue if you get few hundred thousand executions per day. It becomes an issue when you have this running couple of billion times a day (like for us) - especially if there is a better and much less costly way to do it. Converting a list of "exclude" rules to constants and passing them on to Raven is such a way.

test/Sentry/SentryBundle/Test/ErrorTypesParserTest.php
dcramer: yeah they def are

dcramer:
In general I think this is good, and its far easier to configure than other mechanisms.

PR #36, opened 2016-12-29
Give access to all of the RavenClient options

LouisTrezzini:


dcramer:
Not sure why we never took this approach before. I guess to some degree it makes sense to expose top level options that are very common, but its far too complicated to keep up with all of the options sentry/sentry exposes that way.

Thanks!

wachterjohannes:
@dcramer do you have any approach to release current master with this changes?

dcramer:
@wachterjohannes shipped 0.7.0

wachterjohannes:
thanks for fast response 😁

PR #37, opened 2017-01-02
Moving setting userinfo from request-listener to exception-listener

tdel:
Use case : 
 - using AuthorizationChecker::isGranted() trigger a query (I know, it shouldn't but some devs are doing wrong things sometimes...)
 - triggering a query means "hello doctrine" and fetching the result of that query
 - lost severals milliseconds because of that.

In fact, this is done on each request but it should only be done when an exception arise.

This PR move some codes where it should be handled properly.

### DIFF
--- a/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
+++ b/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
@@ -63,20 +63,10 @@ class ExceptionListener
     }
 
     /**
-     * Set the username from the security context by listening on core.request
-     *
-     * @param GetResponseEvent $event
+     * Set the username from the security context
      */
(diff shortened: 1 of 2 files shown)


dcramer:
This regresses behavior in which calling out to capture any kind of handled exception would no longer have the current user set. The goal of context is that you can bind it (albeit, I agree, lazily would be nicer), and then any future events generated within a transaction contain that context. Now it's only doing it after one of two kinds of events happen.

Jean85:
I'm sorry, but I completely disagree on this.

If, as an example, the exception being thrown is somewhat DB-related, the query may not be executable, and this will fail.

Also, if any user is logged in, the query for the user will be executed anyway for each request, so you're just doing it early, you will not save any time for your request.

PR #40, opened 2017-01-26
Update services.yml

szepczynski:
Not quoting the scalar "%sentry.options%" starting with the "%" indicator character is deprecated since Symfony 3.1 and will throw a ParseException in 4.0


dcramer:
thanks!

PR #53, opened 2017-04-27
Log Sentry errors

### LINKED ISSUE #51: Log Sentry errors with Monolog
https://docs.sentry.io/clients/php/usage/#handling-failures - it would be good to know when and why event was not reported to Sentry

duxet:
Resolves #51

### DIFF
--- a/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
+++ b/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
@@ -2,6 +2,7 @@
 
 namespace Sentry\SentryBundle\EventListener;
 
+use Psr\Log\LoggerInterface;
 use Sentry\SentryBundle;
 use Sentry\SentryBundle\Event\SentryUserContextEvent;
 use Sentry\SentryBundle\SentrySymfonyClient;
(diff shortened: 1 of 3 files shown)


Addvilz:
Looks good, but please, add tests in appropriate places.

duxet:
@Addvilz done :)

src/Sentry/SentryBundle/EventListener/ExceptionListener.php
```diff
@@ -63,6 +66,14 @@ public function setClient(\Raven_Client $client)
     }
 
     /**
+    * @param \Psr\Log\LoggerInterface $logger
+    */
+    public function setLogger($logger)
```
Jean85: Please typehint against the interface: `(LoggerInterface $logger = null)`

duxet: I was sure that referencing class that might be not available will throw an exception, but i have checked that now and seems like this is not going to happen :)

Addvilz:
@Jean85 perhaps leaving this disabled by default and adding a configuration option to provide logger service could be better? You can configure and pass whatever service that implements logger interface.

Jean85:
Yes @Addvilz that is a good solution too.

duxet:
I am working on some configuration options, i just need to figure out how it can be done.

Jean85:
Maye you can use https://github.com/facile-it/sentry-psr-log ?

PR #54, opened 2017-05-05
Improve CI

Jean85:
- add PHP 7.1
 - do not allow failure for 7.0
 - cache only composer downloads, not info about lib versions
 - prefer to install dist version of dependencies
 - let the build fail for tests first, CS latter

I would also like to add some external quality/test coverage check tool, but that would require some intervention on the repo settings. Any preference? I like Scrutinizer, but we also have CodeClimate and many others... @dcramer it's up to you, I will gladly add the integration in this PR!

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,22 @@
+checks:
+  php: true
+
+coding_style:
+  php:
+    spaces:
+      around_operators:
+        concatenation: true
+        negation: true
+      other:
+        after_type_cast: false
+
+tools:
+  external_code_coverage: true
+
+  php_code_sniffer:
+    config:
+      standard: "PSR2"
+
+filter:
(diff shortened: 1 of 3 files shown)


dcramer:
I think we were using Scrutinizer on the sentry-php repo, so probably good to be uniform

Jean85:
Ok great! I've added a minimal `.scrutinizer.yml` config and the code coverage integration, but I don't have the privileges to set it up on Scrutinizer; if you can activate the repo there it should be enough!

dcramer:
just setup scrutinizer for this repo

Jean85:
Great, thanks! I've excluded the examples folder from the analysis, and this latest push proved that the PR integration is working correctly! :+1: 

I'll merge this and later I will try to tackle some of the stuff that emerged from this.

PR #58, opened 2017-05-18
test against different dependency versions

xabbuh:
* by default, run with latest stable versions
* add a job with lowest stable constrained deps
* run with upcoming releases of required deps

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,13 +14,17 @@
     "require": {
         "php": ">=5.3.3",
         "sentry/sentry": ">=1.5.0",
-        "symfony/framework-bundle": ">=2.4.0"
+        "symfony/config": "^2.4|^3.0",
+        "symfony/console": "^2.4|^3.0",
+        "symfony/dependency-injection": "^2.4|^3.0",
+        "symfony/event-dispatcher": "^2.4|^3.0",
+        "symfony/http-kernel": "^2.4|^3.0",
+        "symfony/security-core": "^2.4|^3.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^1.8.0",
         "phpunit/phpunit": "^4.6.6"
     },
-    "minimum-stability": "dev",
     "autoload": {
         "psr-0" : {
             "Sentry\\SentryBundle\\": "src/"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -14,12 +14,18 @@
...
+        "symfony/dependency-injection": "^2.4|^3.0|^4.0",
+        "symfony/event-dispatcher": "^2.4|^3.0|^4.0",
+        "symfony/http-kernel": "^2.4|^3.0|^4.0",
+        "symfony/security-core": "^2.4|^3.0|^4.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^1.8.0",
         "phpunit/phpunit": "^4.6.6"
     },
+    "minimum-stability": "dev",
```
Jean85: I don't like the addition of this. I know that this is needed for testing against Symfony 4, but in this way it will allow the installation of a lot of dependencies as dev-master, which could cause a lot of unrelated failures in the build (see [this build](https://travis-ci.org/getsentry/sentry-symfony/jobs/233671298#L207)).

Could we allow SF 4.0 in some other way, like `"^2.4|^3.0|^4.0|^4.0-dev"`?

xabbuh: Sure, that would work. Though I think it would be a good idea to run a dedicated job with dev versions of all required dependencies. Hopefully, this helps spotting issues before new releases of required packages are released. Also note that by default no untagged deps will be installed due to the usage of the `--prefer-stable` option but only one job is leveraging this setting. However, if you are not happy with this change, I will of course update the required version instead. :)

Jean85: Yes I will prefer that, thanks. I've noticed that the `--prefer-stable` avoids untagged releases, but the issue that I was highlighting is on the matrix element that you added, where that flag is not used. Also, declaring that this lib is 4.0-compatible is kind of a lie right now, since it's not released and it can change and have BC at any time now. And if we tag a release with this in the `composer.json` it could create problems in the future, with Composer rolling back to that version because of that.

xabbuh: Alright, I removed the `^4.0` part of the constraint and removed the `minimum-stability` setting alongside the unstable build job.

Jean85:
Thanks @xabbuh! I've pushed an other commit on top to simplify the options (`--prefer-stable` everywhere) and to test with `--prefer-lowest` in 5.3 too (the real lowest of this bundle).

PR #61, opened 2017-06-15
Made priorities configurable

### LINKED ISSUE #49: Listener priority on kernel.exception
Hi, 

Today, i get the exception on sentry even if I catch it from kernel.exception and want to redirect my user instead of seeing it on sentry. 

Can the listener have lower priority on kernel.exception event ? 

What do you think ?

rdohms:
In order to allow users to move the listeners in their priorities queue
the value is now configurable via configuration parameters. In order to
maintain BC these values default to 0 which is the current value.

this closes #49.

### DIFF
--- a/src/Sentry/SentryBundle/DependencyInjection/Configuration.php
+++ b/src/Sentry/SentryBundle/DependencyInjection/Configuration.php
@@ -71,6 +71,14 @@ class Configuration implements ConfigurationInterface
                         '%kernel.root_dir%/../var/cache',
                     ))
                 ->end()
+                ->arrayNode('listener_priorities')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('request')->defaultValue(0)->end()
+                        ->scalarNode('kernel_exception')->defaultValue(0)->end()
+                        ->scalarNode('console_exception')->defaultValue(0)->end()
+                    ->end()
+                ->end()
             ->end()
         ;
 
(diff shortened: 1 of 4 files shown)


rdohms:
Worth noting, I added each parameter as a scalar node to avoid adding complexity to the Extension, but i'm happy to make it an array node and extend the extension if desired..

rdohms:
Fixed the extra line left behind, oops.

Jean85:
First, thanks for the PR!
IMHO, an array with fixed keys would be more elegant. Something like this:
```yaml
listener_priorities:
    request: 0
    kernel_exception: 0
    console_exception: 0
```

Is it too much complicated do be done?

rdohms:
@Jean85 should not be, i had done it initially. I'll add some more complexity to the extension but should be good. Gimme a sec.

rdohms:
@Jean85 there you go. I tried to go with making the `sentry.listener_priorities` parameter an array and then using Symfony expressions in the `services.yml` file, but that has proven to not be supported in the `priority` section.

So i went with the simple approach of making each a scalar parameter, making config nice and clean and leaving the rest hidden behind the bundle wall. 

Let me know if this needs cleaning up more, i decided to not exclude the `sentry.listener_priorities`, but for no strong reason at all.

Jean85:
Thanks, it's perfect! :+1: 
I'll work on the release ASAP

Jean85:
Released: https://github.com/getsentry/sentry-symfony/releases/tag/0.8.0

rdohms:
Thanks!

PR #67, opened 2017-07-24
Listens on console.command to trigger client load

### LINKED ISSUE #66: Error Handler not registered for CLI Commands
Registration of the Client happens when the `Client::install` command is called by the container as it instantiates the Service.

In the scenario of a web request, this is done by the Kernel as it always triggers the `kernel.request` which causes it to load the listener and thus boot the client which then hooks into the PHP Erro Handler and registers itself.

However in the context for a CLI Command, there is no Kernel. And the listener only listens to `console.exception`. Since Symfony lazy-ly only instantiates the listeners if the event happens, when no exception is thrown the listener will never be instantiated, thus the client won't either and will never hook itself into the PHP Error Handler, and thus `PHP Notices` for example, will go by unchecked and never reported to Sentry.

Solutions:
- Add a listener to an event that is always thrown that forces instantiation of the client.
- Add this same command to the current listener
- Find another way to hook into kickoff procedure to ensure the client is *always* instantiated, like  doing this in the Extension for example.

I'm going to explore a few to see if I can find the best way.

rdohms:
Closes #66. 

This hook up forces the boot of the client to happen always when a command is triggered, which before only happened if an exception was thrown.

This should offer no BC issues and the only side effect is catching more notices then before as these may have been going unnoticed.

### DIFF
--- a/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
+++ b/src/Sentry/SentryBundle/EventListener/ExceptionListener.php
@@ -6,6 +6,7 @@ use Sentry\SentryBundle;
 use Sentry\SentryBundle\Event\SentryUserContextEvent;
 use Sentry\SentryBundle\SentrySymfonyClient;
 use Sentry\SentryBundle\SentrySymfonyEvents;
+use Symfony\Component\Console\Event\ConsoleCommandEvent;
 use Symfony\Component\Console\Event\ConsoleExceptionEvent;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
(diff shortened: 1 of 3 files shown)


Jean85:
Thanks for the contribution @rdohms !! I'll try to release this in a patch ASAP.

lcobucci:
@Jean85 I think that something weird happened, this commit is not part of `v0.8.1`: https://github.com/getsentry/sentry-symfony/commits/0.8.1

Jean85:
@lcobucci @rdohms my bad, I handled badly the fact that this PR was targeting master and not the 0.8.x branch. I've now cherrypicked the commit and tagged a 0.8.2 release wich should be fixed.

lcobucci:
@Jean85 no worries, it happens. Thanks!

Issue #41, opened 2017-02-10
Options config value accepts only scalar values


ste93cry:
I was trying to set the `options` config key in the Symfony configuration but it accepts only scalar values while it should accept an array of key/values. Changing the `prototype` of the node from `scalar` to `variable` in the config tree builder works, although I think it would be better to have a full definition that checks all options available so that a user cannot add whatever he wants. Let me know what you think about this, I could submit a PR that fixes the issue

Jean85:
You are right! It should accept a full array, but maybe we should limit the amount possible stuff that can be passed in there?

PR is more than welcome!

ste93cry:
>but maybe we should limit the amount possible stuff that can be passed in there?

Indeed, but does it makes sense to invest time in doing it now when 2.0 version of the base library is being developed?

>It should accept a full array

Indeed, at the moment noone can use the options because of this

michaelzangerle:
I discovered the same issue when setting tags via options and took a look at the configuration tree of the bundle. There seems to be same issue multiple times (options, skip_capture, prefixes, excluded_app_paths, ...). The `prototype` is defined as `scalar` although the default is an `array`.

I thought about creating a PR to define the options that are also used by the Raven-Client but then again some values are set via setter injection and some via the configuration. I can think of three possible solutions:

- A. Add all the options the client accepts to the options property of the configuration tree and get rid of those that would be duplicate like `environment` (BC break).
- B. Create all the configuration options the client accepts at the same level as `environment` and pass the whole configuration on to the client (not that clean). `options` could then be removed.
- C. Leave the existing configuration properties as they are and add the not existing ones to the `options` property (not very clean but would not cause a BC break).

What do you think? I would prefer solution A as it is the cleanest one but some feedback before starting on the PR would be appreciated.

Jean85:
Since `options` is not working in the current form, I wouldn't consider it a BC changing the way it works; I also like option A, but I would prefer to deprecate first and drop later (in next major) the duplicates; we can accept both and throw an exception when **both are defined but with different values**.

In fact this is a mixed A-C approach, but just as a deprecation path towards a full-A case. I think that would be better, since in next major we will probably rewrite the option part anyway, due to the v2 ov the Sentry client.

michaelzangerle:
@Jean85 Do you know when the next major will be released? I am just wondering if it's even worth to change it or just move on to the next version...

Jean85:
[...]

PR #68, opened 2017-08-04
[BUGFIX] fixes the configuration tree

michaelzangerle:
Fixes issues discovered and discussed in issue https://github.com/getsentry/sentry-symfony/issues/41.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,66 +54,19 @@ class AppKernel extends Kernel
 
 ### Step 3: Configure the SDK
 
-Add your DSN to ``app/config/config.yml``:
-
-```yaml
-
-sentry:
-    dsn: "https://public:secret@sentry.example.com/1"
-```
-
-## Configuration
-
-The following can be configured via ``app/config/config.yml``:
-
-### app_path
-
-The base path to your application. Used to trim prefixes and mark frames as part of your application.
-
-```yaml
-sentry:
-    app_path: "/path/to/myapp"
-```
-
-### dsn
-
-[Sentry DSN](https://docs.sentry.io/quickstart/#configure-the-dsn) value of your project.
- Leaving this value empty will effectively disable Sentry reporting.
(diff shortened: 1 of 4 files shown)


michaelzangerle:
@Jean85 The failing tests are the same as in the master branch.

Jean85:
Hi @michaelzangerle, thanks a lot for this contibution!

Master branch [is green](https://travis-ci.org/getsentry/sentry-symfony/builds/258193235), to what failures are you referring to?

Also, I would prefer to not remove the deprecated option from the readme. Can you add a paragraph called "Deprecated options" that just summarizes with a YAML snippet the deprecated structure of option?

Thanks!

michaelzangerle:
@Jean85 Whoops it redirected me to the current build instead of the master build that's why I thought it also failed on master. But still do you have any idea why it's failing? It's not even in this bundle ... 

```
PHP Fatal error:  Class 'Raven_Processor_SanitizeDataProcessor' not found in /home/travis/build/getsentry/sentry-symfony/vendor/sentry/sentry/lib/Raven/Client.php on line 286
```

Yes I can add the deprecated options to the readme but is there any value to it? They should not be used anyway ...

Jean85:
I've restarted the latest build on master and it's still green. You should issue a `composer update --prefer-lowest` to try and reproduce the issue in your local environment.

As for the deprecated options in the README, I would like to write about that to leave a notice about the deprecation itself, something like this:

> ## Deprecated options
> In previous releases of this bundle, some of the previous options where set outside of the `options` level of the configuration file. Those still work but are deprecated, and they will be dropped in the 2.x release, so **you are advised to abandon them**; to provide forward compatibility, they can be used alongside the standard syntax, but values must match. This is a list of those option:
> ```
> sentry:
>     deprecated-option1: ~
>     deprecated-option2: ~
>     deprecated-option3: ~
>     ...
> ```

michaelzangerle:
@Jean85 Done and tests pass now.

PR #71, opened 2017-08-18
Add PHP 7.2 support

Jean85:
~~This is a very small, self-explanatory PR...~~

This PR adds PHP 7.2 in the build matrix and removes the only triggered deprecation because of that.


Jean85:
Build failures seem related to fixed bugs in PHP, since I do not get them running tests onto PHP 7.2.0beta2.

PR #72, opened 2017-08-24
Fix readme options error_types that expected array

JellyBellyDev:
see title of PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,8 @@ Define which error types should be reported.
 ```yaml
 sentry:
     options:
-        error_types: E_ALL & ~E_DEPRECATED & ~E_NOTICE
+        error_types: 
+          - E_ALL & ~E_DEPRECATED & ~E_NOTICE
 ```
 
 #### exception_listener


Jean85:
Well, that was unexpected, thanks for noticing!

I assume you got that because you needed it and found the discrepancy: can you first report back if this seems to work? 
I don't like that it is like this, but changing this now would mean a BC; your PR could be still valid, but I would like to change it in the stable release.

JellyBellyDev:
In the previous release version (before you move options and set deprecated level config), this parameter should have a string value, in the new version it wants it as an array.

Maybe the bug is [here](https://github.com/getsentry/sentry-symfony/blob/c2493d76ee1856133c0214e1a1f48d51c2c0a6e0/src/Sentry/SentryBundle/DependencyInjection/Configuration.php#L76) and not should continue to be a array but a string?

JellyBellyDev:
However changing the configuration as reported in my readme works!

Jean85:
Well, then it's a bug!! I'll fix it right away, thanks for spotting it!

JellyBellyDev:
Happy to have been of help!

Jean85:
I've released a patch: https://github.com/getsentry/sentry-symfony/releases/tag/0.8.6
You're highly advised on updating to it; I'll get on working on the stable release ASAP, which will drop all this options-related nonsense.

JellyBellyDev:
🚀

Issue #79, opened 2017-10-19
No proper error page in case of fatal error when this bundle is activated


hectorj:
Hi!

I'm having a small problem with this bundle:

When I activate it, Symfony doesn't handle PHP fatal errors anymore.

Let's say I have a `$this->nonExistingMethod();` in one of my controller. Without this bundle, in prod env, Symfony displays a "pretty" HTML error page.

With this bundle, I get this instead:

![error 500](https://user-images.githubusercontent.com/2427959/31766668-8c66e0b6-b4c8-11e7-9942-d4740e1944fc.png)

My env is:
PHP 7.1
Symfony v3.3.10
sentry/sentry-symfony 0.8.6

Jean85:
I was able to reproduce it on our prod env, sorry for the issue :( I'll try to debug it...

hectorj:
From what I see, [here](https://github.com/symfony/symfony/blob/v3.3.10/src/Symfony/Component/HttpKernel/EventListener/DebugHandlersListener.php#L118) we get a `\Raven_ErrorHandler` instead of a `\Symfony\Component\Debug\ErrorHandler`, so `\Symfony\Component\HttpKernel\HttpKernel::terminateWithException` (instantiated [here](https://github.com/symfony/symfony/blob/v3.3.10/src/Symfony/Component/HttpKernel/EventListener/DebugHandlersListener.php#L105) and usually registered [here](https://github.com/symfony/symfony/blob/v3.3.10/src/Symfony/Component/HttpKernel/EventListener/DebugHandlersListener.php#L123)) is never registered.

That's because `\set_exception_handler` is called by `\Raven_ErrorHandler::registerExceptionHandler` which is called by `\Raven_Client::install` which is called when Symfony instantiates it (following the config [here](https://github.com/getsentry/sentry-symfony/blob/0.8.6/src/Sentry/SentryBundle/Resources/config/services.yml#L6)), which happens when Symfony instantiates `\Sentry\SentryBundle\EventListener\ExceptionListener`, which happens when Symfony gets all the listeners for the `KernelRequest` event. (sorry for the run-on sentence).

Without the Sentry bundle, `\Symfony\Component\Debug\ErrorHandler` is the registered exception handler and `\Symfony\Component\HttpKernel\HttpKernel::terminateWithException` gets registered into it.

I'm still investigating more.

hectorj:
[...]

Issue #48, opened 2017-04-07
Ignore ErrorException thrown by Symfony


duxet:
After catching errors (like `TypeError`) Symfony is also throwing `ErrorException` and that results in two exceptions being reported to Sentry instead of reporting only the first one.

The simplest solution would be adding `ErrorException` to list of ignored exceptions, but i am not sure if that's a good idea.

delamart:
I was wondering the same and it looks like ignoring ErrorException is not an option.
I had an ErrorException triggered by an undefined array index which I believe is an E_NOTICE so I got a sentry issue for that which I would not have had at all if I ignored ErrorException.

This is all default configuration for symfony and sentry by the way.

Jean85:
This is starting to seem an unrelated issue or something linked to the base PHP SDK. I'm removing this issue from the stable milestone (see #86).

stayallive:
With sentry-php version 1.8.1 just released, this should be fixed (run composer update). If it still persists for anyone, please open an new issue on getsentry/sentry-php with some info on the error that is duplicated.

Jean85:
Closing as this is fixed with `sentry/sentry` 1.8.1.

PR #80, opened 2017-10-20
Dirty workaround for issue #79

hectorj:
I don't think this should be merged as-is, it's just to show a working workaround for issue #79

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -6,6 +6,7 @@ use Sentry\SentryBundle\Event\SentryUserContextEvent;
 use Sentry\SentryBundle\SentrySymfonyEvents;
 use Symfony\Component\Console\Event\ConsoleCommandEvent;
 use Symfony\Component\Console\Event\ConsoleExceptionEvent;
+use Symfony\Component\Debug\ErrorHandler;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
@@ -80,6 +81,8 @@ class ExceptionListener implements SentryExceptionListenerInterface
         if (null === $this->tokenStorage || null === $this->authorizationChecker) {
             return;
         }
+        $this->client->install();
+        ErrorHandler::register();
 
         $token = $this->tokenStorage->getToken();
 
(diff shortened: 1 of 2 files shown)


Jean85:
Thanks for this PR, is a good starting point!
So, basically you're suggesting re-registering the Symfony error handler after Sentry's. Did you check if Sentry's error reporting still works after this workaround?

hectorj:
Yep. According to my manual testing, I got both the Symfony error page and the error logged by Sentry.

hectorj:
> So, basically you're suggesting re-registering the Symfony error handler after Sentry's.

You also need to `install()` the client only after Symfony's `DebugHandlersListener::configure` ran...

I'm not sure how to get there properly.

Jean85:
I'm investigating on it. I've discovered that:

 * `DebugHandlersListener::configure` is called as an event subscriber on events `kernel.request` and `console.command` with priority 2048; registering on the same events using a lower priority should be enough
 * `ErrorHandler::register` is done at the start of the `FrameworkBundle::boot` function; maybe forcing the bundle to be loaded before it is a feasible solution?

hectorj:
> registering on the same events using a lower priority should be enough

Sadly no. With https://github.com/getsentry/sentry-symfony/blob/008479adca17bbe1a6c29d63f9bfdc470e83f073/src/Resources/config/services.yml#L5-L6, `install()` is called when the listener is instantiated, which is before any events is processed.

That's why in this PR I had to move the `install()` call.

Jean85:
Yes yes, I was referring exactly to that, moving the `install()` inside a listener on those events instead that on instantiation.

That's the easy part, what I'm worried about now is the interaction with Symfony's ErrorHandler, that also converts errors into exceptions, and I think is the root cause of #48

hectorj:
Now that I'm thinking about it, do we really need to do the `install()` ? Symfony already takes care of catching fatal errors, exceptions, etc. with its own ErrorHandler.

I think adding Sentry as a monolog handler would do the job properly, or am I missing something?

Jean85:
Installing the client is done to catch non-halting issues, like notices, warnings and deprecations.

hectorj:
So I was indeed missing something... still, the exception handler doesn't seem to be needed in the context of Symfony.

Jean85:
Why? I want to have all the things reported in Sentry....

Jean85:
I tried the proposed solution in my prod env, still got the blank page, it doesn't seem to work.

Anyhow, as stated previously, the issue resides in the interferences with the Symfony's ErrorHandler... I'm closing this and going back to discuss this in #79

PR #82, opened 2017-10-26
fire PRE Capture event before use it

bartosz-foremniak:
It's more logical, in some cases for example in other listener we want to modify exception in event, therefore it should not be skipped and ignored completely.

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -96,13 +96,14 @@ class ExceptionListener implements SentryExceptionListenerInterface
      */
     public function onKernelException(GetResponseForExceptionEvent $event)
     {
+        $this->eventDispatcher->dispatch(SentrySymfonyEvents::PRE_CAPTURE, $event);
+        
         $exception = $event->getException();
 
         if ($this->shouldExceptionCaptureBeSkipped($exception)) {
             return;
         }
-
-        $this->eventDispatcher->dispatch(SentrySymfonyEvents::PRE_CAPTURE, $event);
+       
         $this->client->captureException($exception);
     }
 
@@ -124,6 +125,8 @@ class ExceptionListener implements SentryExceptionListenerInterface
      */
     public function onConsoleException(ConsoleExceptionEvent $event)
     {
+        $this->eventDispatcher->dispatch(SentrySymfonyEvents::PRE_CAPTURE, $event);
+        
         $command = $event->getCommand();
         $exception = $event->getException();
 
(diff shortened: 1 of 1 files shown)


Jean85:
Hi, and thanks for the contribution.

This is a potential BC, so I would allow it only on the master branch, I will not backport it to 0.8. Anyhow, I'm not completely convinced, since in this way a potential listener of the `PRE_CAPTURE` event wouldn't have a way to know if the exception will be skipped afterwards.

Also, tests are broken.

bartosz-foremniak:
> I'm not completely convinced, since in this way a potential listener of the PRE_CAPTURE event wouldn't have a way to know if the exception will be skipped afterwards. 

But if  exception is change in  pre_capture listener it's never check in fact. (It can be fix by use clone before check)

Jean85:
I'm sorry, I didn't understand your comment. The `shouldExceptionCaptureBeSkipped` check is done anyway, with or without your PR; your code changes only that you can alter the exception before the check (using `$event->setException()` at `PRE_CAPTURE`) and change the flow of execution.

But I'm not sure why would you want to do something like that.

bartosz-foremniak:
Before my PR you can alter the exception but **after**  `shouldExceptionCaptureBeSkipped` check -> my PR fix that (exception is pass by reference).
Possible solution for that is to clone exception before fire `PRE_CAPTURE` event, but then we lose possibility to modify it.

Jean85:
But my point is: why would you want to alter it? What's the need for it?

bartosz-foremniak:
It's long story... I maybe prepare proper pull request for these but at the moment i prepare temporary workaround in my project (by extending `Sentry\SentryBundle\EventListener\ExceptionListener`).  

It's really frustrating me that when i throw Exception from my domain layer and it's capture by 
`\Sentry\SentryBundle\EventListener\ExceptionListener::onKernelException` there is no possibilty to send extra params to sentry which will be nice.

So i want to do something like that
```
    /**
     * @param GetResponseForExceptionEvent $event
     */
    public function onKernelException(GetResponseForExceptionEvent $event)
    {
       $this->eventDispatcher->dispatch(SentrySymfonyEvents::PRE_CAPTURE, $event);

        $exception = $event->getException();
        
       if ($this->shouldExceptionCaptureBeSkipped($exception)) {
            return;
        }

        $data = [];

        if ($exception instanceof SomeSentryInterface) {
            $data['extra'] = $exception->getData();
        }

        $this->client->captureException($exception, $data);
    } 
```
It's could be nice feature, but it's still frustrate me, beacuse i don't like the idea that my domain knew something about sentry - the solution is to write event listener that wrap every exception that extends my `DomainException` in something that that implements `SomeSentryInterface`.

The best place for that is `PRE_CAPTURE` event...
So i started wrote that and found that bug (in my opinion) - not skip exception after skip. So i made PR.

I know that i could prepare custom ExceptionListener   (but i need that solution in more than one project) or fork the lib... But it could be nice core feature. 


Sorry for my bad english ;)

Jean85:
Mmmh... according to your use case, I think you should use the `processor` option: it lets you add a class that processes the event before sending it, it should solve your needs!

Jean85:
Closing this as stale. Feel free to reply.

PR #83, opened 2017-10-26
documentation evet_listener fix

bartosz-foremniak:
In Configuration event_listener is under root node.


Jean85:
You're right! Thanks for spotting the issue.

Issue #86, opened 2017-10-31
Create new tagged release


ruudk:
Would it be possible to get a new release out of the door?

`52 commits to master since this release`

The bundle doesn't work out of the box with [Symfony Flex](https://github.com/symfony/recipes-contrib/pull/40) because it wants to get a tagged release with the Symfony Flex fixes.

Jean85:
Hello there! Currently, the master branch contains a very different release (see #73): it should be the next 1.0.

Currenty I was waiting on #43 #48 and #79 to be solved before, but it seems that they are only blocking the stable release without any good enough  reason.

I'll work to prepare the relase ASAP.

ruudk:
@Jean85 Thank you so much! I just started this Recipe on Symfony Flex for 0.8 release: https://github.com/symfony/recipes-contrib/pull/128

Should I wait with this one, and use 1.0 ? :)

Jean85:
Yes, you should! I'm opening a PR right now, I hope to release 1.0 ASAP.

Jean85:
I'm currently thinking about one thing though... Flex is for Symfony 4, but can be used with 3.3+; the bundle is not currently ready to be used with 4.0, we will need a new (major?) release to do that if there are major deprecations in play.

We will need to look into that when 3.4/4.0 will be released in the next 30 days.

ruudk:
What happens when you use this bundle in 4.0? Any known issues?

Jean85:
Right now it will just not install, because any Symfony constraint in `composer.json` says `"^2.7|^3.0"`. 

I intend to bump it to `"^3.0|^4.0"`, but I will be forced to release an other new version to do that; I'm also thinking about bumping the minimum PHP version with that, so we will be able to leverage the new language features, but with both those changes, I think we would need a new (2.0) major release.

ruudk:
If this bundle works with 4.0, we could include it in the v1.0 version right?

ruudk:
Then 2.0 could be used for bumping the PHP version to 7.1 ;)

Jean85:
I would avoid that; allowing more than 2 major versions is too much, and I would like to avoid dropping 2.7 since [it's still actively supported](https://symfony.com/roadmap?version=2.7#checker) and I would like to have at least one stable release that supports that.

I plan to work on the 2.0 ASAP anyway.

Jean85:
Just released 1.0! https://twitter.com/AlessandroLai/status/927816508352466944 :tada:

PR #87, opened 2017-10-31
Preparation to release 1.0

Jean85:
We should now be ready to release this bundle as stable, so we can also have a Symfony Flex recipe; see #86 and https://github.com/symfony/recipes-contrib/pull/128.

I've also upped the constraint to the Sentry SDK to `^1.8` which was released yesterday.

Please @dcramer approve this so we can merge, tag and release!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ cache:
 before_install:
   - composer self-update
   - composer global require hirak/prestissimo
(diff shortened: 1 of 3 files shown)


Jean85:
The build is awfully slow with the `--prefer-lowest` job. I'll try to add `hirak/prestissimo` to speed that up.

Issue #60, opened 2017-05-31
Should you pass command params to exception, too?


andrewtch:
Currently sentry stores only command name as tags. Should you also store command params for later debugging?

dcramer:
We definitely wouldn't want params as tags. Do we actually store the command as a tag, or extra context? I'd expect it to be context. We also wouldn't want to do this by default, as people often pass sensitive information as arguments.

andrewtch:
Yeah, you wouldn't :) but still, feel free to add this as a recipe to add input as extra:

```php
<?php

namespace AppBundle\Sentry;

use Sentry\SentryBundle\SentrySymfonyEvents;
use Symfony\Component\Console\Event\ConsoleExceptionEvent;
use Symfony\Component\EventDispatcher\Event;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;

class EventSubscriber implements EventSubscriberInterface
{
    /**
     * @var \Raven_Client
     */
    protected $client;

    /**
     * EventSubscriber constructor.
     * @param \Raven_Client $client
     */
    public function __construct(\Raven_Client $client)
    {
        $this->client = $client;
    }


    public function onPreCapture(Event $event)
    {
        if ($event instanceof ConsoleExceptionEvent) {
            $this->client->extra_data = [
                'arguments' => $event->getInput()->getArguments(),
                'options' => $event->getInput()->getOptions()
            ];
        }
    }
    
    public static function getSubscribedEvents()
    {
        return [
            SentrySymfonyEvents::PRE_CAPTURE => 'onPreCapture'
        ];
    }
}

```

Jean85:
The command is saved as a tag, currently, and I think is a good practice, since different command means different action.

As per the arguments, I agree that we shouldn't report them by default to avoid logging sensitive data; maybe we can enable it optionally?

andrewtch:
I would just add a recipe to README (see above), like "example event listener: record console exception arguments".

Jean85:
There's already [an (empty) example in the readme](https://github.com/getsentry/sentry-symfony#add-an-eventsubscriber-for-sentry-events), happy to accept a PR if you want to replace it with your working example.

Jean85:
Closing this as stale. Feel free to reply or open a PR if willing.

PR #97, opened 2017-11-23
Add more information about command

bartosz-foremniak:
It's pretty useful to have more information about command parameters


Jean85:
I don't think this should be a default behavior, see https://github.com/getsentry/sentry-symfony/issues/60#issuecomment-305269876

Issue #104, opened 2018-01-30
Console Exceptions being logged twice


artureiji:
Hello, this month my team received a notice that our logging limit was exceeded, and investigating it we found out some errors were being sent to sentry twice.

After a fast test it seems that the Sentry ExceptionListener is listening to both `ConsoleErrorEvent` and `ConsoleExceptionEvent`, that are dispatched by `Symfony\Component\Console\Application` for a single exception that occurs.

sentry-symfony: 2.0
sentry: 1.8.2
symfony: 3.4.4

Jean85:
Thanks for reporting this. I attached the listener to both events because I wanted to have cross compatibility for Symfony 3 and 4, but this is leading to this bug.

I'll try to find an other solution.

Jean85:
Fix released as [2.0.1](https://github.com/getsentry/sentry-symfony/releases/tag/2.0.1)!

xthiago:
Thank you @Jean85

PR #105, opened 2018-01-31
Avoid double-reporting of CLI errors

Jean85:
This should solve #104. Since we already require `symfony/console` 3.3+, we can rely on the `console.error` event only, without the need of the older, deprecated `console.exception`.


Jean85:
I tagged this anyway to speed up the patch release.

Issue #110, opened 2018-02-27
Reducing build time for the flag "--prefer-lowest"


agentsib:
Hello!

At the moment, build time for cases with  "-prefer-lower" flag is extremely slow. In some cases because of the timeout build is unsuccessful.

Just a suggestion. I think the build time can be reduced if you add a file `composer-5.6-lowest.lock`. Before build with '--prefer-lowest' flag you can replace `composer.lock` file and run `composer install`.

Or you can cache file between builds (`composer-5.6-lowest.lock`)

What do you think about it?

PS. Sorry for my english :)

Jean85:
Caching it between builds could be an good idea! But we need a successful build first....
Or maybe we could run a generic `composer install` first and then run it with the `--prefer-lowest`?

agentsib:
> Or maybe we could run a generic composer install first and then run it with the --prefer-lowest?

Without composer.lock file it will not help.

Case 1:

```bash
# time composer install

real	0m5,371s
user	0m2,239s
sys	0m0,514s

# time composer  update --no-interaction --prefer-dist --prefer-stable --prefer-lowest 

real	2m23,637s
user	2m22,082s
sys	0m0,651s

```

Case 2 (https://gist.github.com/agentsib/2e57aacd37a31f2cba8b5273a66d233b):

``` bash
# cp ./build/composer-5.6-lowest.lock ./composer.lock
# composer install 


real	0m1,197s
user	0m0,513s
sys	0m0,383s

# time composer update --no-interaction --prefer-dist --prefer-stable --prefer-lowest 

real	0m16,212s
user	0m12,744s
sys	0m0,088s

**Comment by Jean85** [comment]
Mh, thanks for the detailed test! HAve you tried with the lock file and doing the update directly, without the install?

**Comment by agentsib** [comment]
@ Jean85, without `composer install` slowly. Only copying `*.lock` file and `composer install` before updating helps.

```bash
# rm -rf vendor/ composer.lock 
# cp ./build/composer-5.6-lowest.lock composer.lock
# time composer update --no-interaction --prefer-dist --prefer-stable --prefer-lowest

real	3m52,679s
user	0m0,024s
sys	0m0,011s
```
```bash
# rm -rf vendor/ composer.lock 
# cp ./build/composer-5.6-lowest.lock composer.lock
# time composer install

real	0m1,498s
user	0m0,481s
sys	0m0,362s
[...]

PR #112, opened 2018-02-27
Fix travis slow builds

agentsib:
For improve perfomance we need:

1. Remove package `friendsofphp/php-cs-fixer` before UnitTests
2. Return package on `cs` task

See #110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,10 @@ before_install:
   - composer self-update
   - composer global require hirak/prestissimo
 
-install: travis_retry travis_wait composer update --no-interaction --prefer-dist --prefer-stable $COMPOSER_OPTIONS
+install:
+  - composer install --prefer-dist
+  - composer remove friendsofphp/php-cs-fixer --dev
+  - travis_retry travis_wait composer update --no-interaction --prefer-dist --prefer-stable $COMPOSER_OPTIONS
 
 script:
   - vendor/bin/phpunit -v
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,7 +17,10 @@ before_install:
   - composer self-update
   - composer global require hirak/prestissimo
 
-install: travis_retry travis_wait composer update --no-interaction --prefer-dist --prefer-stable $COMPOSER_OPTIONS
+install:
+  - composer install --prefer-dist
```
Jean85: Is this `composer install` pre-step still required?

.travis.yml
```diff
@@ -17,7 +17,10 @@ before_install:
...
+  - composer remove friendsofphp/php-cs-fixer --dev
```
Jean85: I would move this `composer remove` into the before install step, with a `--no-update` flag to avoid any actual modification of the vendor folder, it should be faster

.travis.yml
Jean85: Is this `composer install` pre-step still required?

.travis.yml
agentsib: `composer remove` will launch `composer install` after delete row from file. Please, see last build.

Jean85: If you use the `--no-update` flag it will not.

agentsib:
@Jean85 Do I still have to do something on the task? Or are you considering another way to solve the problem?

Thanks

PR #116, opened 2018-03-11
add symfony default tags

hjanuschka:
adds `symfony_version` and `symfony_app_env` tags, per default.

### DIFF
--- a/src/SentrySymfonyClient.php
+++ b/src/SentrySymfonyClient.php
@@ -4,7 +4,7 @@ namespace Sentry\SentryBundle;
 
 class SentrySymfonyClient extends \Raven_Client
 {
-    public function __construct(?string $dsn = null, array $options = [], $env = 'undefined')
+    public function __construct(?string $dsn = null, array $options = [])
     {
         if (! empty($options['error_types'])) {
             $exParser = new ErrorTypesParser($options['error_types']);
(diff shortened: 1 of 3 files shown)


Jean85:
Hi there, and thanks for the contribution! You should know though that the Symfony environment by default is already sent to Sentry using the configuration key ``, see here:

https://github.com/getsentry/sentry-symfony/blob/ea4de1359e77127784c58d4dcbc865d148c8a8cb/src/DependencyInjection/Configuration.php#L67

Also, you should simplify the `array_merge` logic using the `??` null coalesce operator.

hjanuschka:
@Jean85 feedback addressed.

ohhh - you are right about the env, somehow in our production system wie set `sentry.environment`  to something custom, this lead me to the mistake that the symfony env is missing.

let me know if you need me to change anything else.

src/SentrySymfonyClient.php
```diff
@@ -15,6 +15,11 @@ public function __construct(?string $dsn = null, array $options = [])
             'name' => 'sentry-symfony',
             'version' => SentryBundle::getVersion(),
         ];
+        $default_tags = [
+          'symfony_version' => \Symfony\Component\HttpKernel\Kernel::VERSION,
+        ];
+        $options['tags'] = $options['tags'] ?? [];
```
Jean85: Umh, I don't understand why you need to do those two lines. Isn't `$options['tags']['symfony_version'] = ...` just enough?

Jean85: The var is initialized as `array $options = []` so it's an array for sure, worst case an empty one.

Since it's an array, setting directly `$options['tags']['symfony_version']` will never fail, so merging is unnecessary.

hjanuschka:
@Jean85 ok now i got you 👍 sorry for me being that dumb, and many thx for patience and guidance. 🍻

Jean85:
Don't worry, thank you for your patience!

Jean85:
Can I ask you to backport this PR to the 1.x branch?

hjanuschka:
@Jean85 sure will submit a seperate PR

PR #118, opened 2018-03-14
add php_version

hjanuschka:
add `php_version` as a default tag.
i upgraded my prod to 7.2 and there where strange errors comming in, and then i realized it was a single node still having <php7.2


Jean85:
Thanks for this contribution, but I think this would be more appropriate in the base client.
BTW I need to investigate if there is a standard tag name that should be used for this purpose.

hjanuschka:
you mean in the sentry-php? you are right, if you want me to do it, i can go for it!
sentry is such a important part in all of our products!


i saw the php_version in the wp-sentry plugin. but yes, it would be best fit in the php client itself.

Jean85:
Yes please, procede with a PR on that repo! We can discuss there where this info should be put.

Thanks, closing here!

PR #131, opened 2018-05-21
Retrieve user IP from request context

### LINKED ISSUE #127: Setting the user IP (with trusted proxies)
As reported in https://github.com/getsentry/sentry-php/pull/585, when using trusted proxies the Sentry SDK doesn't get the right user IP address. We should retrieve this info from the framework.

eliecharra:
Fix #127 

- [x] Fix tests
- [x] Add test

Tested under docker env with symfony configured like this

```yaml
framework:
    trusted_proxies:
        - 172.0.0.0/8 # Docker ip ranges
```

selection_763

Context dumped from symfony

selection_764


Jean85:
Hello, and thanks for this PR! It seems great, can you open a second one from the same branch, but targeting the `releases/1.x` branch, so we can backport this?

Issue #120, opened 2018-03-21
Enabling sentry-symfony hides deprecation messages


Epskampie:
I noticed that enabling this bundle hides symfony deprecation messages.

Steps to reproduce:

* Start a new symfony project (i used symfony 3) `$ symfony new test lts`
* Add a simple deprecated call to `DefaultController`: `$this->getDoctrine()->getEntityManager();`
* Notice there is a deprecation message for this in the web profiler toolbar.
* Now open `AppKernel.php`, and add the SentryBundle (for all environments, not just `prod`)
* Notice the deprecation message is now gone.

I know now that I shouldn't add the SentryBundle in a `dev` environment, but still it would be better if sentry didn't eat those deprecation logs if possible.

ragusa87:
I can confirm this bug using:
- symfony 2.8.40, sentry/sentry-symfony 1.0.2, symfony/monolog-bundle 2.12.1
- symfony 3.4.10, sentry/sentry-symfony 2.0.2, symfony/monolog-bundle 2.12.1

Loading the bundle hide all deprecated messages.
 ```@trigger_error("Test deprecated warning", E_USER_DEPRECATED);```

Jean85:
Sorry that I missed this issue!

Yes, it's not good to use this bundle in dev; also, the issue is not in the bundle in itself but in the underlying SDK; the Sentry client register itself as an error handler, which is probably the same way that Symfony uses to catch the deprecations... so that's probably why we have this bug.

I'm not sure that there is a fix for it... Can you confirm the bug using `sentry/sentry` latest version, 1.9?

ragusa87:
Hello,
Yes, I can confirm that the bug is still there on the latest version of sentry/sentry 1.9 or with dev-master (7bf56a8abc)
I can also confirm you that the raven error handler seems to be the problem, once it is registered, the deprecated messages are hidden (\Raven_ErrorHandler::registerExceptionHandler).
Should we open an issue on the sentry/sentry repository ?

Jean85:
Yes you should, but I don't think you'll find an immediate solution... We should debug how the Symfony handler works and why our handler doesn't pass stuff to him.

ragusa87:
Unfortunately I do not have time to take a deeper look at this yet. However, I opened an issue on sentry: https://github.com/getsentry/sentry-php/issues/605

Thanks for your feedback, and have a nice day.

ragusa87:
[...]

PR #133, opened 2018-05-28
Fix install

### LINKED ISSUE #63: Symfony error pages are not being show correctly
After I enable sentry, when a error 500 happens, the error page is not shown correctly. When I'm in 'dev' environment, the symfony debug page is not shown, as stated in the following images:

Bad:
bad

Good:
good

And when I'm in 'prod' environment, my customized error page is also not shown, as stated in the following images:

Bad:
bad2

Good: 
good2

ragusa87:
Call Raven_Client::install via a listener instead of directly on container boot.
This way, Symfony error handlers are registered before the raven ones, avoiding a few bad behaviors.

Fix #63 #120 

This branch need testing & tagging.


ragusa87:
@Jean85 Can you take a look to my PR ?
I used another listener to call the Raven_Client install method, and I created a custom configuration to be able to change the listener priorities. I also added two tests.

I made a test project here if needed: https://github.com/ragusa87/sentry-notify, you can adapt the composer.json to use various version of sentry-symfony bundle and see the various behaviors.

Jean85:
Hi @ragusa87, thanks for your work! I'll try to investigate further ASAP, but I fear that lowering the priority for that install will hinder the ability of reporting errors... I'll try to use your app to debug this further.

Were you able to reproduce the issue within your test project, before using this fix?

ragusa87:
Hello @Jean85 , I was able to reproduce the issue with my test project before the fix. Anyway it's currently using my fork so you won't see the problem if you keep the sentry-symfony dependency like this. You have to adapt the composer file to use the official bundle version. 

Removing the install call seems to be a quite better approach. So I think it's better to investigate this solution and close the PR.

Jean85:
Yes, I agree. We can continue discussing this in #63.

tmbdrogba:
+1

PR #141, opened 2018-06-10
[Feature] Allow configure sentry.client service as lazy

grachevko:
Sometimes ago i get error while cache warmup inside `docker build`.

Some service was try to instantiate `sentry.client` with this config
```yaml
sentry:
    dsn: '%env(SENTRY_DSN)%'
```
And inside `docker build` there is no environments. 

But after i was create this PR i can't reproduce this error. Maybe migrating to symfony 4.1 solve that. 

If you think this changes not necessary, please close it.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -44,6 +44,9 @@ class Configuration implements ConfigurationInterface
                         ->then($this->getTrimClosure())
                     ->end()
                     ->defaultNull()
+                ->end()
+                ->booleanNode('lazy')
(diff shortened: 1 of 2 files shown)


Jean85:
Making the service lazy may break some internals of this bundle, like including user info on the request. Sorry, but I don't think I should implement this.

If you need to do a docker build, you should provide empty env variables, even in the parameters.yml if you prefer.

grachevko:
@Jean85 to solve this i was write:
```php
$container->getDefinition('sentry.client')->setLazy(true);
```
But everything work as expected.

Issue #143, opened 2018-06-23
Send session data


zhil:
It would be nice to see sesson data in sentry issues. I plan to implement it, could such a PR be accepted?

Jean85:
Session data may frequently be sensitive, so I don't think it should be added as a default.

zhil:
@Jean85 ok, what about disabled by default + some easy way to enable it in config?

Jean85:
That would be better...

zhil:
@Jean85 added #144 
let me know if its acceptable

Jean85:
First and foremost, we should differentiate between `$_SESSION` and the session service, it's not the same.
Second, the sentry-php behavior shouldn't be changed, too much stuff relies on that; the bundle should adapt itself to that, not the other way around.
Third, I think we should clarify what stuff should or shouldn't be sent from the session to Sentry; I fear that sending the whole session data is detrimental at the end.

zhil:
@Jean85 well, current sentry-symfony-bundle + sentry-php behavior is kind of strange - for not logged in users it send all session data(including symfony internals session data), for logged in users - do not send any session data.

as for me - most logical way would be mimic sentry-php behavior, with username instead of session_id and _SESSION array in $user['data']

zhil:
ok, in my case I have such a bug with bundle - session data is not send for logged in users.
I have solved it using simple EventSubscriber
```
<?php
namespace App\SomeBundle\EventSubscriber;

use Sentry\SentryBundle\SentrySymfonyEvents;
use Symfony\Component\EventDispatcher\Event;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\HttpFoundation\Session\Session;
use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;

class MySentryEventSubscriber implements EventSubscriberInterface
{
    /** @var \Raven_Client */
    protected $client;

    /** @var Session */
    protected $session;

    public function __construct(\Raven_Client $client, Session $session)
    {
        $this->client = $client;
        $this->session = $session;
    }

    public static function getSubscribedEvents()
    {
        // return the subscribed events, their methods and priorities
        return array(
            SentrySymfonyEvents::PRE_CAPTURE => 'onPreCapture',
        );
    }


    public function onPreCapture(Event $event)
    {
        if ($event instanceof GetResponseForExceptionEvent) {
            $this->client->user_context($this->session->all(), true);
        }
    }
}
```

[...]

PR #144, opened 2018-06-27
#143 - added send_session_data option

zhil:


src/EventListener/ExceptionListener.php
```diff
@@ -54,13 +61,15 @@ public function __construct(
         EventDispatcherInterface $dispatcher,
         RequestStack $requestStack,
         array $skipCapture,
+        $sendSessionData,
```
Jean85: Please add it to the PHPDoc

src/EventListener/ExceptionListener.php
```diff
@@ -188,4 +203,16 @@ private function setUserValue($user)
             $this->client->set_user_data((string)$user, null, $data);
         }
     }
+
+    private function setSessionData()
+    {
+        if ($this->session) {
+            $this->client->user_context($this->session->all());
```
Jean85: Can you show me an example of how this data is treated inside Sentry?

zhil: image

Jean85:
Build seems to be broken due to PHPStan. I'll try to fix it elsewhere.

Jean85:
Triggering rebuild after merging #145...

PR #145, opened 2018-06-27
Update to PHPStan 0.10 and PHPUnit 7

Jean85:


stayallive:
Why is `/Cannot cast object to string\./` ignored?

Jean85:
Error is reported here: https://github.com/getsentry/sentry-symfony/blob/ee83edf11d36dadf42ab59984c8c69ab94387351/src/EventListener/ExceptionListener.php#L209

PHPStan cannot recognize the `__toString` existence.

Jean85:
PHPStan isn't able to recognize the `method_exists` check here: 

https://github.com/getsentry/sentry-symfony/blob/ee83edf11d36dadf42ab59984c8c69ab94387351/src/EventListener/ExceptionListener.php#L208-L210

See https://github.com/phpstan/phpstan/issues/323#issuecomment-400005174

stayallive:
Cool! I understand :)

PR #154, opened 2018-08-06
Remove ignore from PHPStan

Jean85:
This will fix the build.


Jean85:
This seems no longer necessary...

PR #155, opened 2018-08-20
Move to XML services and remove Yaml dependency

Pierstoval:
I don't know why, but we had a bug on our project on production while upgrading to v2 of the bundle:

`Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 1 passed to Sentry\SentryBundle\EventListener\ExceptionListener::__construct() must be an instance of Raven_Client, instance of Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage given, called in /app/var/cache/prod/ContainerPhwac6d/appProdProjectContainer.php on line 1867 in /app/vendor/sentry/sentry-symfony/src/EventListener/ExceptionListener.php:54
Stack trace:
#0 /app/var/cache/prod/ContainerPhwac6d/appProdProjectContainer.php(1867): Sentry\SentryBundle\EventListener\ExceptionListener->__construct(Object(Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage), Object(Symfony\Component\Security\Core\Authorization\AuthorizationChecker), Object(Sentry\SentryBundle\SentrySymfonyClient), Array, Object(Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher))
#1 /app/var/cache/prod/ContainerPhwac6d/appProdProjectContainer.php(1069): ContainerPhwac6d\appProdProjectContainer->getS in /app/vendor/sentry/sentry-symfony/src/EventListener/ExceptionListener.php on line 54`

I could reproduce this bug locally, and I couldn't find any problem.

First fix I tested was to _name_ all arguments in the `services.yml` config file for the listener's definition, but when I thought about moving from Yaml to XML, everything started to work like a charm… at least locally.

Anyway, XML is better because it can help drop Yaml direct dependency, so let's propose this change here anyway 👍  (I hope I didn't make any mistake while migrationg 🤣 )


Jean85:
This is a nice improvement! Can you solve the conflicts?

src/Resources/config/services.xml
```diff
@@ -0,0 +1,30 @@
...
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="sentry.client" class="%sentry.client%" public="true">
+            <argument>%sentry.dsn%</argument>
+            <argument>%sentry.options%</argument>
+            <call method="install" />
+        </service>
+
+        <service id="%sentry.client%" alias="sentry.client" />
```
Pierstoval: Added this part, hope it's correct :)

Pierstoval:
Done @Jean85 :smile:

src/Resources/config/services.xml
Jean85: Yes it is! :+1:

Issue #157, opened 2018-09-27
Support autowiring for Symfony 4


gander:
I would like to use autowiring in Symfony 4 without having to enter it in the service configuration:

```yaml
services:
    Sentry\SentryBundle\SentrySymfonyClient: '@sentry.client'
```

```php
<?php

namespace App\Controller;

use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
use Sentry\SentryBundle\SentrySymfonyClient;
use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\Routing\Annotation\Route;

class DefaultController extends AbstractController
{
    /**
     * @Route("/", name="homepage")
     * @param SentrySymfonyClient $client
     * @Template()
     */
    public function index(SentrySymfonyClient $client)
    {
    }
}
```

Jean85:
Good idea, PR welcome on this!

Jean85:
This is solved by #158.

PR #158, opened 2018-09-28
Add sentry.client class alias to sentry.client service

gander:
Issue #157


gander:
Hmm, CI test timeout? Can you restart testing?

PR #159, opened 2018-09-28
Update changelog

Jean85:


Jean85:
Ping @HazAT here too, review needed.

Jean85:
Well, that's not reasonable but it seems like @HazAT approval doesn't count... Ping @stayallive for an approval here too.

mcfedr:
2.0.4 tag?

PR #160, opened 2018-10-01
Push to the sentry transaction

mcfedr:
- At start and end of request, using _route attribute if its set
- At start and end of commands, using command name


test/EventListener/ExceptionListenerTest.php
```diff
@@ -510,6 +565,30 @@ public function test_that_it_captures_exception()
         $listener->onKernelException($mockEvent);
     }
 
+    public function test_that_it_sets_transaction_from_command()
+    {
+        $mockEvent = $this->createMock(ConsoleCommandEvent::class);
+
+        $command = new Command();
```
Jean85: This is making the test fail in `--prefer-lowest`; need a name.

src/EventListener/SentryExceptionListenerInterface.php
```diff
@@ -28,6 +31,23 @@ public function onKernelRequest(GetResponseEvent $event): void;
      */
     public function onKernelException(GetResponseForExceptionEvent $event): void;
 
+    /**
+     * When the request finished
+     *
+     * @param FinishRequestEvent $event
+     */
+    public function onFinishRequest(FinishRequestEvent $event): void;
```
Jean85: These new methods  are a huge breaking change :disappointed: I don't think I can merge this right now as it is...

src/EventListener/ExceptionListener.php
```diff
@@ -82,6 +84,10 @@ public function setClient(\Raven_Client $client)
      */
     public function onKernelRequest(GetResponseEvent $event): void
     {
+        if (($request = $event->getRequest()) && ($route = $request->attributes->get('_route'))) {
+            $this->client->transaction->push($route);
```
Jean85: Why do you do a push/pop? Is this for subrequests?
Also, why transaction and not a tag? What's the end result in Sentry?

src/EventListener/ExceptionListener.php
```diff
@@ -138,6 +153,13 @@ public function onConsoleException(ConsoleExceptionEvent $event): void
         $this->handleConsoleError($event);
     }
 
+    public function onFinishCommand(ConsoleTerminateEvent $event): void
+    {
+        if (($command = $event->getCommand())) {
+            $this->client->transaction->pop($command->getName());
```
Jean85: Here too: what's the case for needing to pop a command name?

test/EventListener/ExceptionListenerTest.php
mcfedr: Yes, just looking at how to do this for Symfony 3

src/EventListener/SentryExceptionListenerInterface.php
mcfedr: How is best to do this? I see that currently in the interface it also doesnt have `onConsoleCommand`, but its in the ExceptionListener, maybe this can also just not be added to the interface?

src/EventListener/ExceptionListener.php
mcfedr: `transaction` seems to be appriopriate place to add to the sentry event, according to https://docs.sentry.io/clientdev/attributes/

In `Raven_Client` this is implemented as a stack, not really sure why, but it does actually work nicely with the sub requests pattern, as they will be pushed on, and then removed, putting as back in the context of the main request

src/EventListener/ExceptionListener.php
mcfedr: I mostly put it here for the sake of symetry, again because `Raven_Client->transaction` is a stack.

There is some sense, in that exceptions throw outside of your command would not be marked as related to the command, e.g. if Symfony was to throw an error

mcfedr:
The transaction value gets a special treatment in Sentry UI, 

here `users_list` is my route name in the title area:

screenshot 2018-10-01 at 5 53 42 pm

it is also show amongst the tags:

screenshot 2018-10-01 at 5 53 48 pm

Jean85:
I've changed the target branch due to the fact that #184 will start development of 3.0 on the master branch.

Unfortunately there are conflicts that need to be resolved first. Anyhow, I'm planning on using `Scope`s on 3.0 to push/pop the same stuff.

Jean85:
Just re-implemented the same in #198 for 3.x.

mcfedr:
Closing as this is fixed with #198

PR #161, opened 2018-10-27
Fix deprecation on Symfony 4.2

chalasr:
See https://github.com/symfony/symfony/pull/27476

PR #162, opened 2018-10-29
Fix PHPStan build

Jean85:
As found in #161, we need to fix the build for PHPStan.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         }
     },
     "scripts": {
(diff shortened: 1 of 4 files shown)


Jean85:
Ping @HazAT, I need a review here... alternatively, who should I bother about this stuff in general?

Jean85:
Thanks @stayallive! Nope, you just added a "comment" review, not an "approved" one...

PR #163, opened 2018-10-29
Added NoopExceptionListener that does nothing when an exception happens

adsc-cloudtec:
With this class, the configuration `exception_listener: Sentry\SentryBundle\EventListener\NoopExceptionListener` will disable exceptions being logged by sentry. Uncaught exceptions should still be logged, because they produce a fatal error by default.


Jean85:
This can be closed since the same was done in #247.

PR #166, opened 2018-11-05
Update changelog for 2.1

Jean85:
This is to prepare the relase of 2.1; minor bump instead of patch due to new features and dropping a dep.

I've put tomorrow's date so I can tag it correctly (I hope).


Jean85:
Woha that was fast :smile: I've fixed the release date then.

PR #167, opened 2018-11-07
ExceptionListener > Add `route` tag

ruudk:
This will make it easier to find which controller / route was executed when an exception occurred.

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -111,8 +111,14 @@ class ExceptionListener implements SentryExceptionListenerInterface
             return;
         }
 
+        $data = ['tags' => []];
+        $request = $this->requestStack->getCurrentRequest();
+        if ($request instanceof Request) {
+            $data['tags']['route'] = $request->attributes->get('_route');
+        }
+
         $this->eventDispatcher->dispatch(SentrySymfonyEvents::PRE_CAPTURE, $event);
-        $this->client->captureException($exception);
+        $this->client->captureException($exception, $data);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Jean85:
I'm kinda conflicted on this change. On one hand, I understand how that's useful,  but OTOH I think we are deviating from the correct behavior here, since we could use the `transaction` attribute, see https://docs.sentry.io/clientdev/attributes/

This would change how the errors are logged though...

ruudk:
Can you point me in the right direct where I could add this `transaction` attribute?

Jean85:
From what I can recall, as in https://github.com/getsentry/sentry-php/blob/58172bda837d78de77d6b497f7d372dc2d879115/test/Raven/Tests/ClientTest.php#L822

You can do `$client->transaction->push(...)`

ruudk:
The issue that this PR tries to solve, is that as a developer, it really helps to not only know the URL (which is useful) but also know the route name of the url. It's way easier to copy and paste the route name into your IDE and search for the controller than by looking at the url and trying to match the route in the code.

ruudk:
I'm not sure why Scrutinizer is failing btw, it says:
```
Fixing PHP Coding Style 11:11
Runs coding-style fixes as per your configured coding-style.
```

But it doesn't say what is wrong. I ran `php-cs-fixer` manually and nothing changed.

Jean85:
The conding style on Scrutinizer may be wrong, it's already checked on Travis, it should be disabled... Who has permission to do that? @mitsuhiko? @dcramer?

ruudk:
Any news @Jean85 ?

Jean85:
I do not have permissions on Scrutinizer to fix that. You should add tests to this PR, though.

Jean85:
CI is failing due to CS, please fix it; it can be done automatically with `composer cs-fix`.

ruudk:
Thanks for giving me the command :) Made it super easy to fix!

PR #176, opened 2019-01-02
Releasing scope constraint

bouland:
Issue #175

Issue #175, opened 2019-01-02
Releasing constaint on methods scope in ExceptionListener 


bouland:
As suggested in README

> As a side note, while the above demonstrates a custom exception listener that does not extend anything you could choose to extend the default ExceptionListener and only override the functionality that you want to.

private methods in ExceptionListener should let us override easier way.

Jean85:
Thanks for the contribution, your PR is merged!

PR #180, opened 2019-01-09
Exception listener with dedicated data methods

bouland:
Sorry for my PR #176, i used version 2.0.2 and didn't take last changes.

Issue #175

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -111,14 +111,24 @@ class ExceptionListener implements SentryExceptionListenerInterface
             return;
         }
 
+        $this->eventDispatcher->dispatch(SentrySymfonyEvents::PRE_CAPTURE, $event);
+        $this->client->captureException($exception, $this->getExceptionData());
+    }
+
+    /**
+     * Additional attributes to pass with this event (see Sentry docs).
+     *
+     * @return array
+     */
+    protected function getExceptionData()
+    {
         $data = ['tags' => []];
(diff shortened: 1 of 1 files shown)


Jean85:
I'm sorry but I think I've missed those too... Where did you got those changes? They aren't from this bundle, it seems. Maybe from a fork?

bouland:
> I'm sorry but I think I've missed those too... Where did you got those changes? They aren't from this bundle, it seems. Maybe from a fork?

No fork, just folowing PR 
- [PR 167](https://github.com/getsentry/sentry-symfony/pull/167/files#diff-cdf99235fbbeccfa12764a4be188ae32R114)
- [PR 139](https://github.com/getsentry/sentry-symfony/pull/131/files#diff-cdf99235fbbeccfa12764a4be188ae32R189)

bouland:
My goal is to write this overriding code
```php
class SentryExceptionListener extends ExceptionListener
{
    /**
     * @param object|string|UserInterface|SentryUserInterface $user
     */
    protected function setUserValue($user)
    {
        if (!$user instanceof SentryUserInterface) {
            parent::setUserValue($user);
            return;
        }

        $this->client->set_user_data(
            $user->getId(),
            $user->getEmail(),
            parent::getUserData() + $user->getSentryData()
        );
    }
}
```

bouland:
UP !

What do you think about it ?

Jean85:
Those are breaking changes for anyone that is extending the listener like you, so I cannot merge this as is.

PR #193, opened 2019-02-25
Add before_send option

Jean85:


Jean85:
The build will be probably fixed by #194

PR #199, opened 2019-02-27
Use dependency injection, in the Symfony style

mcfedr:
This is clearly the approach chosen for v3 - I havent been following the development, but using a global static variable instead of DI, just seems like a massive step backwards. Using Symfony I'd expect everything to be using DI, and that if I want to replace the `HubInterface` service, its going to be replaced everywhere its used.

I'm also not going to use a global variable in my own code, I'm going to inject `HubInterface`

### DIFF
--- a/src/EventListener/ConsoleListener.php
+++ b/src/EventListener/ConsoleListener.php
@@ -18,7 +18,7 @@ final class ConsoleListener
      */
     public function __construct(HubInterface $hub)
     {
-        $this->hub = $hub; // not used, needed to trigger instantiation
+        $this->hub = $hub;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Jean85:
I agree that a full-DI approach would be better, but the fact that the `Hub` is a singleton cannot be ignored here, and it comes from the [Unified API](https://docs.sentry.io/development/sdk-dev/unified-api/) approach: it's something defined for all languages supported by Sentry, so there's some trade-off to make it applicable to everyone.

Also, this would break if anyone changed the current hub, because the integration would act on an old hub, not the active/current one. You could call `$this->hub->getCurrent()->...` every time, but that would render this change irrelevant and make the code just harder to decipher.

Issue #156, opened 2018-08-20
Enabling this bundle suppresses errors being logged by symfony and duplicates errors being sent


ostrolucky:
Symfony 3.4.14
sentry-symfony 2.0.3
sentry/sentry 1.9.2

enabling this bundle causes errors being sent twice:
![image](https://user-images.githubusercontent.com/496233/44336929-475bd980-a479-11e8-8764-6330b2adc6b8.png)

It also causes errors not being logged in prod.log

adsc-cloudtec:
To avoid the sentry event duplication, you can simply implement an own ExceptionListener that does nothing. Uncaught exceptions will still be logged as sentry events, because they are fatal errors in PHP.

See also #148

adsc-cloudtec:
I have added such an ExceptionListener class in this PR: #163

Jean85:
This double-reporting may stem from the fact that Sentry is wired through two different things:
 * as an error handler, with `Client::install()`
 * with Symfony's error handler

I'm thinking maybe we should migrate away from the first option, to avoid overlapping like this one. But I don't want to introduce BCs. Maybe we can create an option to disable it?

Jean85:
@ostrolucky I'm working on the 3.0 release (which is currently on the master branch) and I've tried removing the listeners to avoid reporting twice, but I'm not reporting anything, since Symfony is catching all exceptions, so nothing hits the Sentry error handler.

How where you able to report the error twice? I need a reproducing case.

ostrolucky:
I no longer have access to project that was using Sentry.

> Symfony is catching all exceptions, so nothing hits the Sentry error handler.

IIRC symfony error handler doesn't just swallow the error, but passes it to other previously registered error handlers, doesn't it?

Jean85:
Sentry's error handler is registered later, so it's him that should get the error and forward to Symfony's. But to me it seems that everything is caught, so no error/exception handler is involved.

ostrolucky:
@martinlutter can you check if this is still a problem in enrola? you still have access

martinlutter:
@ostrolucky errors are still duplicated in Enrola, and on a different project where I've implemented Sentry.
Also the error suppression is still happening, if Sentry is loaded, the default PHP error logging is suppressed completely. So if Sentry can't report the error, there ais no log entry about it anywhere.
I can try to send the code to reproduce later but the implementation is really basic. Maybe worth mentioning that we use Monolog for logging errors locally.

[...]

PR #204, opened 2019-03-19
Rework to use only listeners

Jean85:
While testing the beta1, I discovered that Sentry's error handler wasn't triggered, if not only on the most fatal errors. Any other intermediate approach lead to double-reporting of errors (like #156), so I have to fall back to using the framework events, and disable Sentry's error handler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## 3.0.0-beta2 - 2019-03-22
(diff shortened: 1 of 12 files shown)


Jean85:
Note to self: is getsentry/sentry-php#788 is merged/tagged, it will break this one. We need to wait that before releasing a stable version.

ste93cry:
@Jean85 if we want to strictly follow semver then that PR will be merged only in `sentry-php 2.1` which will probably see the light not in the near future. I suggest to just merge this PR and rework it later when it will be time. Also I should have implemented everything in a BC-compatible manner, so you shouldn't have the need of adapting your changes to mine... Can you explain to me what is going to break?

Jean85:
It would break in the fact that you will add a new handler, that I would have to remove to avoid double-reporting of fatal errors.

PR #208, opened 2019-04-16
Add info about skip_capture in the update document

### LINKED ISSUE #207: Upgrading to 3 is missing info about options
I'm trying to upgrade to latest beta version.
So, I changed re requirement in my composer from `^2.3` to `^3.0@beta` and ran `composer update`. Then I read the [doc](url).
Now I'm getting an error, since the key `skip_capture` under `options` is not supported anymore.
It looks like mentioned doc is missing that part.

Jean85:
This fixes #207

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -8,8 +8,10 @@ The major change is in the fact that we now require the `sentry/sdk` metapackage
 This new version has been completely rewritten: if you use this bundle and you interact directly with the underlying SDK
 and client, you should read through the [relative upgrade document](https://github.com/getsentry/sentry-php/blob/master/UPGRADE-2.0.md).
 
-The only breaking change reflected in the bundle due to this update is the `skip_capture` option: it has been **removed** because
-the same feature has been implemented in the new SDK, with the native `excluded_exceptions` option.
+## Changes in the configuration
+There are only two BC impacting the configuration:
+ * the `skip_capture` option has been **removed**: the same feature has been implemented in the new SDK, with the native `excluded_exceptions` option; you just have to move your values from `sentry.skip_capture` to `sentry.options.excluded_exceptions`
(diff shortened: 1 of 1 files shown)


garak:
As noted in issue comment, also `curl_method` option needs to be removed when upgrading.

Jean85:
That's a native SDK option, everything there is covered in the [related doc](https://github.com/getsentry/sentry-php/blob/master/UPGRADE-2.0.md).

garak:
@Jean85 but doc says that you need to read related doc only if you "interact directly with the underlying SDK".
People not using such interaction can still have `curl_method` option, so we should tell them that the option needs to be removed.

UPGRADE-3.0.md
```diff
@@ -8,6 +8,11 @@ The major change is in the fact that we now require the `sentry/sdk` metapackage
 This new version has been completely rewritten: if you use this bundle and you interact directly with the underlying SDK
 and client, you should read through the [relative upgrade document](https://github.com/getsentry/sentry-php/blob/master/UPGRADE-2.0.md).
 
+## Changes in the configuration
+There are only two BC impacting the configuration:
+ * the `skip_capture` option has been **removed**: the same feature has been implemented in the new SDK, with the native `excluded_exceptions` option; you just have to move your values from `sentry.skip_capture` to `sentry.options.excluded_exceptions`
+ * The `sentry.options` values reflect the options of the native SDK; many of those have been removed, and there are some new ones. You can read the [appropriate section of the upgrade document](https://github.com/getsentry/sentry-php/blob/master/UPGRADE-2.0.md#client-options), with the exception of the `server` to `dsn` migration, which is still handled in the same way by the bundle, under `sentry.dsn`
```
stayallive: Possibly 'PHP SDK' instead of native.

PR #213, opened 2019-05-15
Fix handling ESI request

franmomu:
Hi!

When the RequestListener is handling a ESI request it fails because it has no `_route` defined in its attributes and it returns `null`, so the `Scope:: setTag` method fails because is expecting a `string`.

### DIFF
--- a/src/EventListener/RequestListener.php
+++ b/src/EventListener/RequestListener.php
@@ -88,7 +88,11 @@ final class RequestListener
             return;
         }
 
-        $matchedRoute = $event->getRequest()->attributes->get('_route');
+        if (! $event->getRequest()->attributes->has('_route')) {
+            return;
+        }
+
+        $matchedRoute = (string) $event->getRequest()->attributes->get('_route');
 
         Hub::getCurrent()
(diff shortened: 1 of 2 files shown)


franmomu:
Sorry, I should have done it in the first place 😬

src/EventListener/RequestListener.php
```diff
@@ -88,6 +88,10 @@ public function onKernelController(FilterControllerEvent $event): void
             return;
         }
 
+        if (! $event->getRequest()->attributes->has('_route')) {
+            return;
+        }
+
         $matchedRoute = $event->getRequest()->attributes->get('_route');
```
franmomu: I was thinking that `$matchRoute` must be a `string` because `Scope::setTag` is typed with `string` and `$event->getRequest()->attributes->get('_route');` returns `mixed`, so maybe there should be a cast in here?

Jean85: Since that depends on Symfony, we can count in it being a string, but a cast doesn't hurt, yeah.

franmomu: done 😬

Issue #217, opened 2019-06-10
Missing info about event deprecations between 2.3 and 3.0


mbabker:
In my app, I had been using the `SentrySymfonyEvents::SET_USER_CONTEXT` event to add my user context info to Sentry.  When updating to 3.0, I found this event is no longer present with no mention of this B/C break in the upgrade guide.  I can work around this, but I think it'd be helpful if the upgrade docs at least noted the `SentrySymfonyEvents::PRE_CAPTURE` and `SentrySymfonyEvents::SET_USER_CONTEXT` events were removed so there's less surprise for users of the bundle.

Jean85:
You're right! As a code example, the basic user data set by the bundle should be enough, we just have to point that in the doc.

Jean85:
I've opened #218, WDYT?

mbabker:
Works for me, thanks!

Stoakes:
Hello,

Do these removals explain why with `sentry-symfony 3.0.0`, instead of having username of affected user I now just have his IP address ?

mbabker:
The default `RequestListener` only processes the IP address, if you need more info you'll need your own listener.  I opted to replace the `RequestListener::onKernelRequest()` listener with my own listener for this purpose:

```php
<?php declare(strict_types=1);

namespace App\EventListener;

use App\Entity\User;
use Sentry\State\Hub;
use Sentry\State\HubInterface;
use Sentry\State\Scope;
use Symfony\Component\HttpKernel\Event\GetResponseEvent;
use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;

final class SentryContextListener
{
    /**
     * @var HubInterface
     */
    private $hub;

    /**
     * @var TokenStorageInterface|null
     */
    private $tokenStorage;

    /**
     * @var AuthorizationCheckerInterface|null
     */
    private $authorizationChecker;

    public function __construct(HubInterface $hub, TokenStorageInterface $tokenStorage, AuthorizationCheckerInterface $authorizationChecker)
    {
        $this->hub                  = $hub; // not used, needed to trigger instantiation (copied from SentryBundle's RequestListener)
        $this->tokenStorage         = $tokenStorage;
        $this->authorizationChecker = $authorizationChecker;
    }

    public function onKernelRequest(GetResponseEvent $event): void
    {
        if (!$event->isMasterRequest()) {
            return;
[...]

PR #220, opened 2019-06-13
Add info about send_default_pii in the upgrade doc

Jean85:
In reference to https://github.com/getsentry/sentry-symfony/issues/217#issuecomment-501712318


untitaker:
@teohhanhui i believe this should work better now, let me know what you think

PR #225, opened 2019-07-04
Exposes Sentry\Monolog\Handler service for easier monolog configuration.

phpeek:
I ❤️ using this bundle.

This PR solves one problem that's still cumbersome to me.

`sentry/sentry` ships with `Sentry\Monolog\Handler` for smooth integration with `monolog`, however this bundle is not exposing it as a service making integration more difficult.

My PR configures `Sentry\Monolog\Handler` as a service, which - in theory - should simplify configuration.

I've updated example for reference, but it's commented out since not everyone might like it configured out-of-the-box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,30 @@ sentry:
     dsn: "https://public:secret@sentry.example.com/1"
 ```
 
+#### Optional: use monolog handler provided by `sentry/sentry`
+*Note: this step is optional*
+
+If You're using `monolog` for logging e.g. in-app errors, You
+can use this handler in order for them to show up in Sentry. 
+
+First, define `Sentry\Monolog\Handler` as a service in `app/config/services.yml`
+
+```yaml
+services:
+    sentry.monolog.handler:
+        class: Sentry\Monolog\Handler
+```
+
+Then enable it in `monolog` config:
+
+```yaml
+monolog:
+    sentry:
+      type: service
+      id: sentry.monolog.handler
+      level: error 
+```
+
 ## Maintained versions
  * 3.x is actively maintained and developed on the master branch, and uses Sentry SDK 2.0;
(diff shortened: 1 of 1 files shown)


phpeek:
Second option of integration is via https://github.com/getsentry/raven-php, but it's marked as deprecated.

phpeek:
Good idea. Will do that later today. Thanks for looking at this !

phpeek:
I've added test You've requested, but it's failing due to a valid reason.

If someone would install this bundle without having `monolog` installed, exposing `Handler` as a service would fail due to lack of the class `Monolog\Handler\AbstractProcessingHandler` that is being extended.

`sentry/sentry` has `monolog/monolog` dependency but it's a `dev` one - which is valid.

Will wait for Your feedback, but I'm leaning towards closing this PR.

phpeek:
What would make sense and probably a better solution is to register this handler as a separate type in `SentryExtension` only when `MonologBundle` is present. That way we could have config like

```
monolog:
   sentry:
       type: sentry
       level: error
```

Not sure if that's worth it. WDYT ?

Jean85:
Surely defining that service every time would be bad for sure. Instead of closing, maybe we can reroute this to a README-only PR? You could suggest the service definition there too.

phpeek:
As suggested, I've updated only `README.md` with optional instructions on how to setup monolog handler.

Jean85:
Thanks!

README.md
```diff
@@ -90,6 +90,30 @@ sentry:
...
+services:
+    sentry.monolog.handler:
+        class: Sentry\Monolog\Handler
+```
+
+Then enable it in `monolog` config:
+
+```yaml
+monolog:
+    sentry:
```
pbowyer: Doesn't this need to be
```yaml
monolog:
    handlers:
        sentry:
```
?

Jean85: Oh right! #230

Issue #235, opened 2019-09-13
Deprecation warnings in Symfony 4.4


garak:
>Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" will return "\Symfony\Component\Config\Definition\Builder\TreeBuilder" as of its next major version. Doing the same in child class "Sentry\SentryBundle\DependencyInjection\Configuration" will be required when upgrading.

Since this repo is already requiring php>7.1, it should be easy to avoid such warning (adding a return type should be enough).
I can provide a PR

Jean85:
Fix seems to be fine, PR is appreciated.

PR #237, opened 2019-09-16
fix deprecated code

garak:
See issue #235

### DIFF
--- a/src/DependencyInjection/SentryExtension.php
+++ b/src/DependencyInjection/SentryExtension.php
@@ -32,7 +32,7 @@ class SentryExtension extends Extension
      *
      * @throws InvalidConfigurationException
      */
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/SentryExtension.php
```diff
@@ -32,7 +32,7 @@ class SentryExtension extends Extension
      *
      * @throws InvalidConfigurationException
      */
-    public function load(array $configs, ContainerBuilder $container)
+    public function load(array $configs, ContainerBuilder $container): void
```
Jean85: Is this another deprecation?

garak: indeed, it isn't.
But it looks weird to get all methods in dir with return types except this one

PR #238, opened 2019-09-17
Fix monolog confiugration for handler and update paths for symfony 4

alexander-schranz:


teohhanhui:
This would still result in duplicate events being sent to Sentry. Some dedupe mechanism is necessary.

Jean85:
It's a duplication because you're actively enabling two reporting paths. IMO they shouldn't be used together, but Monolog should at the very least used as a breadcrumb populator.

PR #244, opened 2019-09-20
Sentry/sentry 2.2 compat

Jean85:


Jean85:
The ZendFramework one was updated too: https://github.com/facile-it/sentry-module/blob/d9d9e6dace1cbb447d88874eb7193aaf20068b1f/composer.json#L48

Issue #248, opened 2019-09-26
Missing option `max_request_body_size` in configuration


JanMikes:
Hello!

This is same situation as with #245. Option `max_request_body_size` is missing in configuration object.

Jean85:
On it! #249 should solve this.

PR #249, opened 2019-09-27
Add max_request_body_size option

Jean85:
This will solve #248.


src/DependencyInjection/Configuration.php
```diff
@@ -170,9 +179,19 @@ private function isNotAValidCallback(): \Closure
...
+    private function maxRequestBodySizeIsSupported(): bool
+    {
+        return $this->optionIsSupported('max_request_body_size', 'none');
+    }
+
+    private function optionIsSupported(string $name, $defaultValue): bool
     {
         try {
-            new Options(['class_serializers' => []]);
+            new Options([$name => $defaultValue]);
```
JanMikes: Do you think this construction is better than `method_exists()` for setter?

Jean85: Performance shouldn't be an issue because this piece of code runs only once during Symfony's cache-clear, but you're right, that would be a better approach. Fixing...

src/DependencyInjection/Configuration.php
```diff
@@ -171,12 +180,11 @@ private function isNotAValidCallback(): \Closure
 
     private function classSerializersAreSupported(): bool
     {
-        try {
-            new Options(['class_serializers' => []]);
+        return method_exists(Options::class, 'getClassSerializers');
```
JanMikes: Imho this is much more readable for developers than the original code, good job here!

Issue #250, opened 2019-09-28
When using Monolog\Handler informations added by sentry listeners are missing


JanMikes:
Hi, there is something very weird going on.

Here is my config
```yaml
services:
    Sentry\Monolog\Handler:
        arguments:
            $hub: '@Sentry\State\HubInterface'
            $level: 'error'

monolog:
    handlers:
        sentry:
            type: service
            id: Sentry\Monolog\Handler
            level: notice
```

But when i log, etc `$this->logger->critical()`, event in sentry does not contain extra informations added by `Sentry\SentryBundle\EventListener\RequestListener`.

By debugging i found out there are 2 different instances of `HubInterface`. 1 lives in our DI container and second on is in static `SentrySdk::$currentHub` property. 

What `RequestListener` does, is only saving this to the static hub, meanwhile in monolog handler, the instance from DI is wired:
<img width="982" alt="Screenshot 2019-09-28 21 05 07" src="https://user-images.githubusercontent.com/3995003/65821437-c2272480-e235-11e9-868a-f9662efd6a1f.png">

On a screenshot you can see these are 2 different instances of hubs, one contains client (created and autowired by DI) the other does not.

I am thinking, if there is not missing `SentrySdk::setCurrentHub()` with the hub created by DI?

Maybe i am just missing something, but this causes many confusions for me.

This might be related to #247

JanMikes:
Basically creating listener like this, before Sentry's one fixes the issue:
```php
public function __construct(HubInterface $hub) {
        $this->hub = $hub; 
    }

    public function onKernelRequest(GetResponseEvent $event): void
    {
        SentryBundle::setCurrentHub($this->hub);
    }
```

JanMikes:
Basically there is completely missing how to use sentry for symfony applications - should i use `\Sentry\captureException();` or inject/autowire `HubInterface` and then `$this->hub->captureException()` as both results in different logs in sentry as described in the issue.

Jean85:
That's exactly what we discussed when we had to implement the Unified APIs in `sentry/sentry`.

I agree that there should be only one hub in the app if the user doesn't create one itself. We should treat this as a bug and try to not create a duplicated hub as a side effect.

Jean85:
[...]

PR #253, opened 2019-09-30
Try to fix Hub bug introduced in #244

Jean85:
See #250.


Jean85:
Fix confirmed to be working: https://github.com/getsentry/sentry-symfony/issues/250#issuecomment-536500434

PR #255, opened 2019-10-02
Added example of ClientBuilder customization to readme

### LINKED ISSUE #251: How to configure ClientBuilder etc with custom serializer?
Hello, i have been struggling about how to set custom serializer and pass it to `ClientBuilderInterface::setSerializer()`.

This is final solution i was able to get working, but i am unsure if it is correct and there is no easier way?

Either i think in config i should use `Sentry\ClientBuilderInterface` instead of `Sentry\ClientBuilder` but since it is private service (defined in services.xml) and when i tried to exchange that i received `Error: Cannot instantiate interface Sentry\ClientBuilderInterface`

```yaml
services:
    Entrydo\Sentry\GenericSentrySerializer: ~
    Entrydo\Sentry\SentryClientBuilderSerializerConfigurator: ~
    Sentry\ClientBuilder:
        configurator: ['@Entrydo\Sentry\SentryClientBuilderSerializerConfigurator', 'configureSerializer']
```
```php
final class SentryClientBuilderSerializerConfigurator
{
    /**
     * @var GenericSentrySerializer
     */
    private $genericSentrySerializer;

    public function __construct(GenericSentrySerializer $genericSentrySerializer)
    {
        $this->genericSentrySerializer = $genericSentrySerializer;
    }


    public function configureSerializer(ClientBuilderInterface $clientBuilder): void
    {
        $clientBuilder->setSerializer(
            $this->genericSentrySerializer
        );

        ClientBuilderConfigurator::configure($clientBuilder);
    }
}
```
```php
final class GenericSentrySerializer implements SerializerInterface
{
    /**
     * @var Serializer
     */
    private $originalSerializer;

    public function __construct(Options $options)
    {
        $this->originalSerializer = new Serializer($options);
    }


    /**
     * Serialize an object (recursively) into something safe to be sent in an Event.
     *
     * @param mixed $value
     *
     * @return string|bool|float|int|object|array|null
     */
    public function serialize($value)
    {
        // and other logic

        if ($value instanceof UuidInterface) {
                return [
                    'uuid' => $value->toString(),
                ];
            }

        return $this->originalSerializer->serialize($value);
    }
}
```

JanMikes:
Closes #251

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,25 @@ default run with a lower priority of `0`.
 Those listeners are `final` so not extendable, but you can look at those to know how to add more information to the 
 current `Scope` and enrich you Sentry events.
 
+
+#### Services configuration
+
+Services registered by the bundle (for example `ClientBuilder`) can be configured in several ways, one of them is by compiler pass:
+
+```php
+final class SentryCustomizationCompilerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $containerBuilder): void
+    {
+        // Example - setting own serializer to Sentry\ClientBuilder
+        $containerBuilder->getDefinition(ClientBuilderInterface::class)
+            ->addMethodCall('setSerializer', [
+                new Reference(MyCustomSerializer::class),
+            ]);
+    }
+}
+```
+
 [Last stable image]: https://poser.pugx.org/sentry/sentry-symfony/version.svg
 [Last unstable image]: https://poser.pugx.org/sentry/sentry-symfony/v/unstable.svg
 [Master build image]: https://travis-ci.org/getsentry/sentry-symfony.svg?branch=master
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -150,6 +150,25 @@ default run with a lower priority of `0`.
 Those listeners are `final` so not extendable, but you can look at those to know how to add more information to the 
 current `Scope` and enrich you Sentry events.
 
+
+#### Services configuration
+
+Services registered by the bundle (for example `ClientBuilder`) can be configured in several ways, one of them is by compiler pass:
```
Jean85:
```suggestion
Services registered by the bundle (the `ClientBuilder` in the example below) can be configured in several ways, one of them is by using a [compiler pass](https://symfony.com/doc/current/service_container/compiler_passes.html):
```

JanMikes:
Thank you, your suggestion makes sense to me.

PR #247, opened 2019-09-25
Options to disable error listener and/or enable Monolog handler

HypeMC:
Resolves #232.

As mentioned in the issue, having both the Monolog handler & the ErrorListener installed results in having duplicate entries.

I added a configuration option to disable the ErrorListener. I also added the ability to configure the Monolog handler through the bundle.

I still need to add some tests as well, but I wanted to get some feedback first on whether or not this approach is good.

PR #260, opened 2019-10-15
Notice of monolog handler available since 3.2.0

Pheetbag:
I added a notice of the monolog handler available since the version 3.2.0  of the bundle to the README.md file.

I install this bundle following the Sentry guide offered in the docs an ended up with the 3.0 version and faced this problem. Finally i figure it out but i think it will be helpful for other developers to add this notice in the README.md


Jean85:
For future reference, the feature was added with #247.

Build is failing for unrelated cause, I'll first fix that.

Issue #264, opened 2019-10-22
RequestListener does not collect user-data


cklm:
There are cases where line 72 in RequestListener fails, even if an user is authenticated (for example, when dealing with more than one firewall). Symfony throws an `Symfony\Component\Security\Core\Exception\AuthenticationCredentialsNotFoundException` in that case. I suggest changing that line to check, if an user is given in $token.

Jean85:
Solved with #265.

PR #265, opened 2019-10-22
Update RequestListener.php

### LINKED ISSUE #263: Undefined variable in RequestListener
Line 77 in RequestListener throws a notice when the if-statement above is false:
`Notice: Undefined variable: userData`

cklm:
fixes #263 and #264

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
  - Fix handling of command with no name on `ConsoleListener` (#261)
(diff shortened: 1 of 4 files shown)


cklm:
Added entries in changelog - we are still using the `AuthorizationChecker` to check, if user-handling is activated. Seems not to be an error for me.

Jean85:
Are you sure? We already have this check on the TokenStorage for that:

https://github.com/getsentry/sentry-symfony/blob/0a71ea681854bbe419bd14953ba8a7aa54d53dc2/src/EventListener/RequestListener.php#L63-L65

cklm:
> Are you sure? We already have this check on the TokenStorage for that:
> 

you are right - thats a double-check - removed it

cklm:
@Jean85 can you have a look here?

Issue #183, opened 2019-01-22
Document the auto-inclusion of HttpExceptionInterface


matt-usurp:
The `HttpExceptionInterface` is not documented anywhere that it is ignored by this bundle. So users who implement the `HttpExceptionInterface` or extend `HttpException` and expect a sentry issue to be raised will be confused.

Having it as the default value in the configuration is also un-intuitive as suddenly `HttpException`s will be raised in sentry when I add my own exceptions to this list.

Unsure how to go about fixing this without just mentioning it as a gimmick in the documentation.

Jean85:
I'm currently working on the 3.0 of this bundle, and I just stumbled on this.

For previous versions, changing that value would create issues for anyone using it implicitly. I've put it there because I think that default avoids capturing 404 and the like; can you confirm that to me?

matt-usurp:
Yeah, the default does avoid capturing the `404` errors, also any other exception that might be implementing that interface. This to me is not ideal as the `HttpExceptionInterface` is there to be implemented and doing so might hide my `CustomSuperFatalException` should I have made one.

I know this is a breaking change so that's why I recommend documenting it here and removing this functionality entirely in the new major version. In the new version document that this should is _recommended to be added_ although not. In the recipe for `symfony/flex` make it part of the default configuration that is installed, this way the developer can remove it or add to the list.

Consider these scenarios:

**Scenario 1** (the base installation)
```yaml
sentry: ~
```
```php
throw new Symfony\Component\HttpKernel\Exception\NotFoundHttpException();
```
The `NotFoundHttpException` will **NOT** be raised in sentry along with all other implementations of `HttpExceptionInterface`.

**Scenario 2** (the problem case)
```yaml
sentry:
  skip_capture:
    - 'App\Exception\CustomException'
```
```php
throw new Symfony\Component\HttpKernel\Exception\NotFoundHttpException();
```
The `NotFoundHttpException` will **NOW** be raised in sentry along with all other implementations of `HttpExceptionInterface` unless I manually add the default value also.

To clarify in **scenario 2** I have added my own exception and now `symfony/config` will no longer apply the default value this bundle provides. This is (in my opinion) a poor developer experience as there is an unknown switch in functionality by simply adding what I would consider to be "another" exception to the list.

TL;DR I don't think there should be any default skipped exceptions.

Jean85:
Closing this as v3 will not include any default exclusions. Feel free to send a PR to the 2.x branch for docs if you want.

PR #269, opened 2019-11-05
Restore ignoring default framework exceptions

rvanlaak:
`sentry-symfony` started relying on the SDK's `excluded_exceptions` implementation, which is great but removes that the SDK is aware of Symfony exceptions that did get caught by default before.

We did not have any exceptions configured in `excluded_exceptions` yet, meaning that since `skip_capture` was removed in v3 the defaults the implementation used to rely on are gone. 

When upgrade to v3, Sentry starts capturing `HttpExceptionInterface` instances.

This PR adds the exceptions Sentry by default does not need to catch. Overriding them is as easy as before, by configuring `excluded_exceptions`.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -217,12 +217,12 @@ class Configuration implements ConfigurationInterface
         return method_exists(Options::class, 'getMaxRequestBodySize');
     }
 
-    private function frameworkDefaultExceptions() : array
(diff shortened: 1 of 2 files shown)


Jean85:
FTR, previous requests on this matter were raised in #183

rvanlaak:
Major versions are there to break stuf. Get the arguments after reading #183 so will close this one. Thank you for the improvement on the docs!

PR #270, opened 2019-11-05
Add note about SF default exceptions

Jean85:


UPGRADE-3.0.md
```diff
@@ -11,6 +11,7 @@ and client, you should read through the [relative upgrade document](https://gith
 ## Changes in the configuration
 There are only two BC impacting the configuration:
  * the `skip_capture` option has been **removed**: the same feature has been implemented in the new SDK, with the PHP `excluded_exceptions` option; you just have to move your values from `sentry.skip_capture` to `sentry.options.excluded_exceptions`
+ * Symfony internal exceptions are no longer ignored by default: this means that you could start getting events for 404 or 403, or even 403 which are followed by a (remembered) login. If you want to ignore those, you will just have to add new values to the `sentry.options.excluded_exceptions` option; notice that it works with an `instanceof` against the exception, so you could ignore multiple kind of events using a commong ancestor class.
```
rvanlaak:
```suggestion
 * Symfony internal exceptions are no longer ignored by default: Sentry will start getting events for 404 or 403, or even 403 which are followed by a (remembered) login. Add new values to the `sentry.options.excluded_exceptions` option; notice that it works with an `instanceof` against the exception, so you could ignore multiple kind of events using a common ancestor class.

rvanlaak:
@Jean85 feel free to merge this simple doc change 👍

PR #266, opened 2019-10-25
Add support for Symfony 5

Big-Shark:

PR #277, opened 2019-11-25
Add basic E2E test

Jean85:
This basic test will fix integration issues with the framework like the one in https://github.com/getsentry/sentry-symfony/pull/266#issuecomment-558054278


Jean85:
Only failure in CI seems to be only when using `sentry/sentry` from develop, due to https://github.com/getsentry/sentry-php/pull/919

I've opened #279 to track that, so we can ignore that failure.

ernestas-s:
Please guys @HazAT or @stayallive

Jean85:
Latest commit broke the build :man_facepalming: 

I think we need to say bye-bye to Symfony 2.8 along with everything < 3.4; they're no longer supported too!

PR #278, opened 2019-11-25
Test against PHP 7.4

Jean85:
This will guarantee support for PHP 7.4, which will be released in only 3 days!


Jean85:
Already done in #282

PR #280, opened 2019-11-28
Replace the use of deprecated getException() by getThrowable()

jeroennoten:
`getException()` on `GetResponseForExceptionEvent` are deprecated since Symfony 4.4.

This PR makes sure that `getThrowable()` is called instead. We use `method_exists()` to check if this method exists, and otherwise fallback to `getException()` to keep support for Symfony 4.3 and before.

### DIFF
--- a/src/EventListener/ErrorListener.php
+++ b/src/EventListener/ErrorListener.php
@@ -23,7 +23,14 @@ final class ErrorListener
 
     public function onKernelException(GetResponseForExceptionEvent $event): void
     {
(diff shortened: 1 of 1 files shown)


Jean85:
This is already done in #277, since I've added the deprecation handler.

Closing as duplicate.

jeroennoten:
@Jean85, do you have any idea when a new version containing these changes will be released? There have been a lot of commits in the past last two months, without any new released version.

PR #283, opened 2019-11-29
Prepare releases 3.2.1

Jean85:
This branch is not to be merged right away, and is not to be deleted afterwards. It detaches the current master, because we need to release a patch (3.2.1) before merging #277, which drops support for older Symfony versions, so we need to release it as the next minor (3.3.0).

So, what we need to do:
 - @HazAT needs to prepare the release by committing the right date in the changelog alongside the 3.2.1 title
 - @HazAT has to tag that same commit as 3.2.1
 - afterwards, I will resolve the conflict of this PR and merge it, but NOT delete the branch, that will stay for eventually further patches on 3.2.x


Jean85:
https://github.com/getsentry/sentry-symfony/releases/tag/3.2.1 released!

I've just fixed the conflicts.

PR #288, opened 2019-12-11
fix PHPUnit 9 deprecations

rvanlaak:
After #266 was merged tests still were failing. Found out this was because of deprecation warnings on PHPUnit. This PR fixes these deprecations on the test suite.

PHPStan is the only test that's failing. What about not running PHPStan on the `/tests` folder?

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit 
+<phpunit
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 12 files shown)


phpunit.xml
```diff
@@ -8,6 +8,9 @@
         cacheResult="false"
         processIsolation="true"
 >
+    <php>
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
```
Jean85: Hard failure on Symfony deprecations is required: we cannot let this bundle start them, or we could flood Sentry with unuseful events.

rvanlaak: I was under the assumption that hereby we would ignore the indirect deprecations. 

Yes, the direct deprecations on this bundle's code should definitely fail CI. Maybe change the mode then?

PR #292, opened 2020-01-02
#291 defaulting `release` config to detected root package version

### LINKED ISSUE #291: Automatically detect application release?
While digging through this integration, I noticed that the docs include a `release:` config.

That config is not set by default, but since this package relies on `ocramius/package-versions`, it may be trivial to set a default, by using `$defaultVersion = \PackageVersions\Versions::getVersion(\PackageVersions\Versions::ROOT_PACKAGE_NAME);`

That is much simpler than requiring the downstream consumers of the package to invent a release/deployment version environment variable.

Ocramius:
Effectively automates release tracking for the grand majority of
applications out there, without having to resort to to any `VERSION`
environment variables nor other custom approaches to be figured out
by the end consumer of the library.

Ref https://docs.sentry.io/workflow/releases/?platform=php

Fixes #291

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
  - Add support for Symfony 5.0 (#266, thanks to @Big-Shark)
  - Drop support for Symfony < 3.4 (#277)
+ - Add default value for the `release` option, using the detected root package version (#291 #292, thanks to @Ocramius)
 
 ## 3.2.1 (2019-12-19)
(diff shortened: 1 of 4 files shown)


Ocramius:
Note: this is a new feature, so `3.3.0` should probably be the target

Ocramius:
> FTR, the new dependency was already there because it's required already by `jean85/pretty-package-versions`

Yes, I added it because it is now explicitly referenced in the sources

Ocramius:
Thanks for merging, @Jean85! :balloon: 

Looking forward to the release \o/

Jean85:
I cannot release directly, I hope to have it done by next week!

PR #293, opened 2020-01-08
Enable PsrLogMessageProcessor on MonologHandler

ruudk:
This will make sure that parameters inside a log record string are interpolated with the values from the context.

Because Sentry reports like this are not very useful:
```
Payment failed with {exception}. Message: {message}
```

Also, everything that is logged as `context` will not be sent to Sentry at all... seems like a different issue.

### DIFF
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -54,10 +54,15 @@
             <tag name="console.command" />
         </service>
 
+        <service id="sentry.monolog.psr_log_message_processor" class="Monolog\Processor\PsrLogMessageProcessor" public="false"/>
+
         <service id="Sentry\Monolog\Handler" class="Sentry\Monolog\Handler" public="false">
             <argument type="service" id="Sentry\State\HubInterface" />
             <argument key="$level" />
             <argument key="$bubble" />
+            <call method="pushProcessor">
+                <argument type="service" id="sentry.monolog.psr_log_message_processor" />
+            </call>
         </service>
     </services>
 </container>


src/Resources/config/services.xml
```diff
@@ -58,6 +58,11 @@
             <argument type="service" id="Sentry\State\HubInterface" />
             <argument key="$level" />
             <argument key="$bubble" />
+            <call method="pushProcessor">
+                <argument>
+                    <service class="Monolog\Processor\PsrLogMessageProcessor" public="false"/>
```
Jean85: This has to be declared as a service somewhere else to work.

ruudk: Yeah, PHPStorm read the XSD and showed this was possible. I was also a bit surprised by this ;) Changed it.

guilliamxavier:
Maybe playing the devil's advocate here, but... What if you *don't* want this processor to be added?

For reference, Symfony MonologBundle adds it by default (except for `type: service` handlers), but allows you to prevent that with the `process_psr_3_messages: false` per-handler option.

As for SentryBundle, currently it doesn't add it by default, but you can add it to the `sentry` handler with
```yaml
services:
    Monolog\Processor\PsrLogMessageProcessor:
        tags: { name: monolog.processor, handler: sentry }
```
(assuming the [current documented configuration](https://github.com/getsentry/sentry-symfony/blob/3.2.1/README.md#optional-use-monolog-handler-provided-by-sentrysentry-available-since-320)).

Now, if it starts adding it by default, how will you be able to prevent that?

ruudk:
@guilliamxavier Wow, that's really clever :) I was not aware that it was possible like this.

I will try this out first, to see if it solves it. If it does, I'll update the docs to explain this to others.

ruudk:
Works, created a PR https://github.com/getsentry/sentry-symfony/pull/295

PR #298, opened 2020-01-14
Support sentry/sentry 2.3 and fix deprecations

Jean85:
This PR fixes the build and supports the 2.3 client correctly, avoiding all deprecations.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "php": "^7.1",
         "jean85/pretty-package-versions": "^1.0",
         "ocramius/package-versions": "^1.3.0",
-        "sentry/sdk": "^2.0",
+        "sentry/sdk": "^2.1",
         "symfony/config": "^3.4||^4.0||^5.0",
         "symfony/console": "^3.4||^4.0||^5.0",
         "symfony/dependency-injection": "^3.4||^4.0||^5.0",
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,7 +6,6 @@
         backupGlobals="false"
         bootstrap="vendor/autoload.php"
         cacheResult="false"
-        processIsolation="true"
 >
     <php>
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -87,7 +87,7 @@ class Configuration implements ConfigurationInterface
             ])
             ->prototype('scalar');
         $optionsChildNodes->arrayNode('excluded_exceptions')
-            ->defaultValue($defaultValues->getExcludedExceptions())
+            ->defaultValue([])
             ->prototype('scalar');
         $optionsChildNodes->scalarNode('http_proxy');
         $optionsChildNodes->arrayNode('integrations')
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -75,6 +75,11 @@ class Configuration implements ConfigurationInterface
             ->defaultValue('%kernel.environment%')
             ->cannotBeEmpty();
         $optionsChildNodes->scalarNode('error_types');
+        $optionsChildNodes->arrayNode('in_app_include')
+            ->defaultValue([
+                '%kernel.project_dir%',
+            ])
+            ->prototype('scalar');
         $optionsChildNodes->arrayNode('in_app_exclude')
             ->defaultValue([
                 '%kernel.cache_dir%',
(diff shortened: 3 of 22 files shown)


composer.json
```diff
@@ -22,7 +22,7 @@
         "php": "^7.1",
         "jean85/pretty-package-versions": "^1.0",
         "ocramius/package-versions": "^1.3.0",
-        "sentry/sdk": "^2.0",
+        "sentry/sdk": "^2.1",
```
Jean85: This locks against `sentry/sentry:^2.3`. Because of this, this should be released as 3.4, to leave space for 3.3.x patches.

phpunit.xml
```diff
@@ -6,7 +6,6 @@
         backupGlobals="false"
         bootstrap="vendor/autoload.php"
         cacheResult="false"
-        processIsolation="true"
```
Jean85: I've used an annotation directly on the only test that needs it, the `End2EndTest`, so now the test suite is faster.

src/DependencyInjection/Configuration.php
```diff
@@ -58,14 +57,10 @@ public function getConfigTreeBuilder(): TreeBuilder
...
             ->ifTrue($this->isNotAValidCallback())
             ->thenInvalid('Expecting callable or service reference, got %s');
-        if (PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion() !== '2.0.0') {
-            $optionsChildNodes->booleanNode('capture_silenced_errors');
-        }
-        if ($this->classSerializersAreSupported()) {
-            $optionsChildNodes->arrayNode('class_serializers')
-                ->defaultValue([])
-                ->prototype('scalar');
-        }
```
Jean85: Here (and below) I've removed all the conditions that were needed for backward compatibility with `sentry/sentry` 2.0, it's no longer needed.

src/DependencyInjection/Configuration.php
```diff
@@ -114,7 +112,7 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->defaultValue($defaultValues->getPrefixes())
             ->prototype('scalar');
         $optionsChildNodes->scalarNode('project_root')
-            ->defaultValue('%kernel.project_dir%');
+            ->setDeprecated();
```
Jean85: This is not a double deprecation, it will pop up during cache warmup, so the user can be notified before reaching production.

src/SentryBundle.php
```diff
@@ -23,24 +22,14 @@ public static function getSdkVersion(): string
...
     {
-        if (class_exists(SentrySdk::class)) {
-            SentrySdk::setCurrentHub($hub);
-
-            return;
-        }
-
-        Hub::setCurrent($hub);
+        SentrySdk::setCurrentHub($hub);
     }
```
Jean85: Those two methods are left here for BC.

src/DependencyInjection/Configuration.php
```diff
@@ -75,14 +70,19 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->defaultValue('%kernel.environment%')
             ->cannotBeEmpty();
         $optionsChildNodes->scalarNode('error_types');
+        $optionsChildNodes->arrayNode('in_app_include')
+            ->defaultValue([
+                '%kernel.project_dir%',
+            ])
```
Jean85: I just noticed that this behavior doesn't work well, and doesn't reproduce the previous behavior, due to https://github.com/getsentry/sentry-php/issues/953 :disappointed:

src/DependencyInjection/Configuration.php
```diff
@@ -75,14 +70,19 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->defaultValue('%kernel.environment%')
             ->cannotBeEmpty();
         $optionsChildNodes->scalarNode('error_types');
+        $optionsChildNodes->arrayNode('in_app_include')
+            ->defaultValue([
+                '%kernel.project_dir%/src',
```
Jean85: Here I'm trying to guess the source folder, since using the parent is not feasibile due to https://github.com/getsentry/sentry-php/issues/953

micheh: @Jean85 With the change in sentry-php 2.3.1 all files are now included by default. Therefore, wouldn't an empty default value make more sense?

With the current default value `'%kernel.project_dir%/src'`, it is impossible to exclude files inside `'%kernel.project_dir%/src'`, since `in_app_include` overrides `in_app_exclude`. With an empty array as default value, this would still be possible.

Jean85: This is true, but it shouldn't have any practical effect you since you can override this default.

stayallive: You still might want to remove it since it takes precedence over `in_app_exclude` which you need to know / remember to clear out the `in_app_include` if you want to exclude anything that is in the `src` dir.

If you keep it empty by default you can just start using the `in_app_exclude` without also needing to clear the `in_app_include` default.

It's not critical ofcourse but it is easier to not have it set and possibly make a `in_app_exclude` value not work because of it.

Jean85: Yup, that's true! Done in #311

PR #308, opened 2020-01-23
Avoid class alias as a BC mechanism

### LINKED ISSUE #307: Class aliasing RequestEvent causes incompatibilities with other bundles that detect Symfony versions by checking class existance
So basically it looks like making a class_alias for RequestEvent for Symfony 5.0 compatibility caused my application to break. 

I have SchebTwoFactorBundle in my app which also needs support for 3.4/4.4/5.0 versions and uses an approach of having a main listener with RequestEvent and a Legacy Listener with GetResponseEvent. It uses the legacy listener when RequestEvent class does not exist (see: https://github.com/scheb/two-factor-bundle/pull/233/files#diff-d45aea21a8ed2417fbab0229857c370aR107). However, when combined with SentryBundle 3.3.0+ this class now exists because it has been class_aliased in two files actually (RequestListener.php and SubRequestListener.php). 

I asked about it on Symfony Slack's #support and several users pointed out that it is the SentryBundle that should fix it's compatibility with different events by doing similar thing to TwoFactorBundle which is to have a legacy listener that uses GetResponseEvent.

And I guess other class aliases I've seen in the PR may break other bundles that detect Symfony versions by checking if a new class exists.

For now I've downgraded the bundle to 3.2.1.

Jean85:
This fixes #307. Using `class_alias` as a BC detection mechanism may break BC for third party packages.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -17,6 +17,12 @@ parameters:
          -
              message: '/Symfony\\Component\\HttpKernel\\Event\\GetResponseForExceptionEvent.$/'
              path: src/EventListener/ErrorListener.php
+         -
+             message: '/(GetResponseEvent|FilterControllerEvent)/'
+             path: src/EventListener/RequestListener.php
+         -
+             message: '/GetResponseEvent/'
+             path: src/EventListener/SubRequestListener.php
 
 includes:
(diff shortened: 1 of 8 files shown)


Jean85:
Closing in favor of #309

PR #281, opened 2019-11-28
Remove deprecated listener events

jderusse:
Fixes https://github.com/getsentry/sentry-symfony/pull/266#issuecomment-558170085

PR #309, opened 2020-01-23
Fix class_alias issues

Jean85:
This is an alternative approach to solve #307; it's a lot simpler than #308, thanks @scheb for implementing it.


jderusse:
reminds me something... https://github.com/getsentry/sentry-symfony/pull/281/files

Jean85:
Yup, you were right..

PR #314, opened 2020-01-29
Use injected Hub instead of SentrySdk::getCurrentHub

### LINKED ISSUE #275: Sentry monolog handler ignores user context configuration
Having configuration in YAML as follows:
```
sentry:
  dsn: "https://xxx@xyz.com"
  register_error_listener: false # Disables the ErrorListener
  options:
    send_default_pii: true
    server_name: xyz.com
  monolog:
    error_handler:
      enabled: true
      level: notice
```

`send_default_pii` seems to be ignored. From what I gather scope created by Handler overwrites the one used by `Hub` or the code related to attaching user data is never called.

Steveb-p:
Fixes #275.

Considering that `HubInterface` is injected anyway, it should be used instead of the global singleton.

### DIFF
--- a/src/EventListener/RequestListener.php
+++ b/src/EventListener/RequestListener.php
@@ -2,7 +2,6 @@
 
 namespace Sentry\SentryBundle\EventListener;
 
-use Sentry\SentrySdk;
 use Sentry\State\HubInterface;
 use Sentry\State\Scope;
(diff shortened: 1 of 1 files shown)


Steveb-p:
@Jean85 thanks, understood :smile:

PR #313, opened 2020-01-28
Remove space in class_alias

starred-gijs:
Introduced in #309

The space in the classname is obviously a typo.

PR #315, opened 2020-01-30
Improve class_alias robustness

guilliamxavier:
- Use `::class` rather than string literal, to avoid problems like #313
- Use distinct alias names (concatenate current file class + external class) rather than `class_exists`, to avoid conflicts without triggering autoload looking for nonexistent files like "src/EventListener/UserContextRequestEvent.php"

### DIFF
--- a/src/EventListener/RequestListener.php
+++ b/src/EventListener/RequestListener.php
@@ -14,19 +14,11 @@ use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInt
 use Symfony\Component\Security\Core\User\UserInterface;
 
 if (Kernel::MAJOR_VERSION >= 5) {
-    if (! class_exists('Sentry\SentryBundle\EventListener\UserContextRequestEvent')) {
-        class_alias(RequestEvent::class, 'Sentry\SentryBundle\EventListener\UserContextRequestEvent');
-    }
-    if (! class_exists('Sentry\SentryBundle\EventListener\UserContextControllerEvent')) {
-        class_alias(ControllerEvent::class, 'Sentry\SentryBundle\EventListener\UserContextControllerEvent');
-    }
+    class_alias(RequestEvent::class, RequestListenerRequestEvent::class);
+    class_alias(ControllerEvent::class, RequestListenerControllerEvent::class);
 } else {
-    if (! class_exists('Sentry\SentryBundle\EventListener\UserContextRequestEvent')) {
-        class_alias(GetResponseEvent::class, 'Sentry\SentryBundle\EventListener\UserContextRequestEvent');
-    }
-    if (! class_exists('Sentry\SentryBundle\EventListener\UserContextControllerEvent')) {
-        class_alias(FilterControllerEvent::class, 'Sentry\SentryBundle\EventListener\UserContextControllerEvent');
-    }
+    class_alias(GetResponseEvent::class, RequestListenerRequestEvent::class);
+    class_alias(FilterControllerEvent::class, RequestListenerControllerEvent::class);
 }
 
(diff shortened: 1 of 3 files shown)


guilliamxavier:
I will try to make PHPStan happy (after meal)...

guilliamxavier:
After searching and trying many things, there appears to be a limitation of PHPStan with `::class` on undefined classes (even with `class_exists` / `class_alias`), e.g. the following snippet:

```php
<?php
namespace My\Ns;
class Foo {}
if (!class_exists(Bar::class)) {
    class_alias(Foo::class, Bar::class);
}
function f(Bar $x): void {
    var_dump($x);
}
f(new Foo());
```

is valid and will output (https://3v4l.org/pPO3n)

```
object(My\Ns\Foo)#1 (0) {
}
```

but PHPStan complains (https://phpstan.org/r/b7443e0a-20b9-4e77-8710-51538dfbea57)

```
+--------------------------------------------------------------------------------+
| Line | test.php                                                                |
+--------------------------------------------------------------------------------+
| 4    | Class My\Ns\Bar not found.                                              |
| 5    | Class My\Ns\Bar not found.                                              |
| 7    | Parameter $x of function My\Ns\f() has invalid typehint type My\Ns\Bar. |
| 10   | Parameter #1 $x of function My\Ns\f expects My\Ns\Bar, My\Ns\Foo given. |
+--------------------------------------------------------------------------------+

 [ERROR] Found 4 errors 
```

So I ended up ignoring the errors of https://travis-ci.org/getsentry/sentry-symfony/jobs/643821562#L644

src/EventListener/RequestListener.php
```diff
@@ -14,19 +14,11 @@
 use Symfony\Component\Security\Core\User\UserInterface;
 
 if (Kernel::MAJOR_VERSION >= 5) {
-    if (! class_exists('Sentry\SentryBundle\EventListener\UserContextRequestEvent')) {
```
Jean85: Removing this `if` may declare the alias twice... I'm not sure if this has any downside.

guilliamxavier: Of course declaring the same alias twice would be no good (https://3v4l.org/7jYDp), that's why I changed the name "UserContextRequestEvent" (copy-pasted from FOSHttpCacheBundle) to distinct names "RequestListenerRequestEvent" in `RequestListener.php` and "SubRequestListenerRequestEvent" in `SubRequestListener.php`, so there's no re-declaration (https://3v4l.org/2OaCD). (Note that these aliases are namespaced, i.e. `\Sentry\SentryBundle\EventListener\RequestListenerRequestEvent`, so another "RequestListenerRequestEvent" can still be declared in another namespace, without causing conflict.)

Now if you fear that "RequestListenerRequestEvent" might itself be declared twice (in the same namespace), the only way I can think of (other than actually declaring the same alias in another file) would be requiring the same file twice (manually, because impossible via autoloading (https://3v4l.org/qPo3m)) but then you would get a fatal error for the class anyway (https://3v4l.org/QqT6H).

Keeping the `class_exists` triggers useless autoloading failures (https://3v4l.org/22IBN), unless adding a `false` 2nd argument (https://3v4l.org/BZYpE), but from the above it seems clear that it would be a useless check. Or do you see something else?

Jean85: Thanks for the explanation, I missed the detail about the two different names :+1:

guilliamxavier: Sorry, it seems that `class_exists` (with `, false` to avoid bound-to-fail autoload) is not always useless after all [when PHP 7.4 preloading enters the game](#354) :s

PR #325, opened 2020-03-16
Improve release default value

### LINKED ISSUE #323: Handle default release version w/o git SHA gracefully
When there is no `.git` data available (e.g. docker setup with only a subfolder of the repo as the shared volume), the version of the root package used for the `release` tag is returned as something like `1.0.0@`.

As suggested in https://github.com/Ocramius/PackageVersions/issues/119, `$version->getPrettyVersion()` of https://github.com/Jean85/pretty-package-versions could be used instead to handle this gracefully.

This is currently not just a "cosmetic" issue, as the sentry frontend seems to not handle this correctly: https://github.com/getsentry/sentry/issues/17474

Jean85:
Fixes #323, improves upon #292. 
Since PrettyVersion is already a requirement of this bundle, we can use it directly to obtain a prettier release tag from the project.

Note: this will still not work if the `.git` folder is missing.


src/DependencyInjection/Configuration.php
```diff
@@ -111,7 +112,7 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->prototype('scalar');
         $optionsChildNodes->scalarNode('project_root');
         $optionsChildNodes->scalarNode('release')
-            ->defaultValue(Versions::getVersion(Versions::ROOT_PACKAGE_NAME))
+            ->defaultValue(PrettyVersions::getVersion(Versions::ROOT_PACKAGE_NAME)->getPrettyVersion())
```
Ocramius: Is this still going to lead to trackable hashes in sentry? The previous version kinda did

Jean85: This is the behavior, from the readme:
> if the requested package is a tagged version, it will return just the short version; if not, [...] it will return the version of the package, followed by the short version of the commit hash (i.e. `6.0.0@fa5711`)

So, it should.

Ocramius: @Jean85 most software is released via rolling `master` or such directly, so it will likely be `dev-master@abcdef`

If sentry can work with that, we should be OK, just wanted to make sure of that.

composer.json
```diff
@@ -21,7 +21,6 @@
     "require": {
         "php": "^7.1",
         "jean85/pretty-package-versions": "^1.0",
-        "ocramius/package-versions": "^1.3.0",
```
guilliamxavier: This removal seems wrong, as the code is still depending directly on `\PackageVersions\Versions::ROOT_PACKAGE_NAME` in `src/DependencyInjection/Configuration.php`

Jean85: That's right, a composer-checker run would have spotted that. BTW this will not break the app since that is an implicit dependency of `jean85/pretty-package-versions`.

Issue #259, opened 2019-10-10
Sentry does not send events to server in messenger context


bjalt:
The sentry/sentry/src/Transport/HttpTransport.php:81 will not send events while running in a worker process. Switching the statement to ``$promise = $this->httpClient->sendRequest($request);`` will work. 
One can also observer how the ``$this->pendingRequests[]`` is holding all the pending request. But non is getting send to the sentry server for display in the sentry frontend.

Can You please advice how to continue?

bjalt:
Found there is already an issue open about this: https://github.com/getsentry/sentry-php/issues/811

dpinheiro:
Although the referenced issue has been closed still I can't get the messenger errors to show up on sentry. Any update to this?

Jean85:
https://github.com/getsentry/sentry-php/pull/813 was released with the 2.1.2 client. You could try to call `flush()` when the messenger dispatches some event that marks the end of a single message?

If that fixes it, I could accept a PR implementing the same, like an additional listener.


Also: do you receive any message if the worker shuts down? Like, if you set it up to shut down every N messages?

dpinheiro:
Sorry for the late reply. Not sure about you mean about the `flush()`, at least I'm not aware (except the doctrine one)

I just tried to set messenger to shutdown after X seconds but no messages sent, I will try to debug the problem

Jean85:
Ok. If no message is being sent at all, it may be that you have something else broken. I would suggest you to use the `sentry:test` command to see if sending a simple event from CLI works at all.

Feel free to open a different issue if you need it.

emarref:
I have found that subscribing to the `WorkerMessageFailedEvent` event is a good way to do it. E.g. https://gist.github.com/emarref/d8f7c2e071e334b39a5042b06927f881

Is it worth creating a pull request adding this subscriber if the application is using Messenger?

Jean85:
Yes it is!

PR #326, opened 2020-03-18
Capture messenger exceptions

emarref:
Errors that occur when handling Messenger messages asynchronously (e.g. background queue) are not captured or flushed to Sentry. This pull request addresses two issues;

Firstly, exceptions thrown during the execution of the Worker class are caught and logged, and therefore never captured by Sentry.

Secondly, captured messages are never flushed to Sentry unless the worker is configured to run with a limit (i.e. `--time-limit=3600` or `--limit=10`).

When a handler fails when executed via the Worker, Messenger dispatches an event. This pull request listens for this `WorkerMessageFailedEvent`, and captures the exception, then flushes immediately.

Addresses #259 and #811.


composer.json
```diff
@@ -34,12 +34,13 @@
...
         "monolog/monolog": "^1.11||^2.0",
         "php-http/mock-client": "^1.0",
-        "phpstan/phpstan-shim": "^0.11",
         "phpstan/phpstan-phpunit": "^0.11",
+        "phpstan/phpstan-shim": "^0.11",
         "phpunit/phpunit": "^7.5||^8.5",
         "symfony/browser-kit": "^3.4||^4.0||^5.0",
         "symfony/expression-language": "^3.4||^4.0||^5.0",
         "symfony/framework-bundle": "^3.4||^4.0||^5.0",
+        "symfony/messenger": "^4.3||^5.0",
```
Jean85: This requirement needs to be ignored in the CI job where we require Symfony 3.4 (and hope it doesn't collide :crossed_fingers: )

src/EventListener/MessengerListener.php
```diff
@@ -0,0 +1,47 @@
...
+        $this->client = $client;
+    }
+
+    /**
+     * @param WorkerMessageFailedEvent $event
+     */
+    public function onWorkerMessageFailed(WorkerMessageFailedEvent $event): void
+    {
+        if ($event->willRetry()) {
+            // Only capture the hard fails. I.e. not those that will be scheduled for retry.
```
Jean85: I'm not sure about this. We probably want to receive errors for those too, or they could fall under the radar.
Otherwise I would make this optional (but not the default).

src/EventListener/MessengerListener.php
```diff
@@ -0,0 +1,47 @@
...
+        if ($error instanceof HandlerFailedException) {
+            // Unwrap the messenger exception to get the original error
+            $error = $error->getPrevious();
+        }
+
+        $this->client->captureException($error);
+
+        // Flush normally happens at shutdown... which only happens in the worker if it is run with a lifecycle limit
+        // such as --time=X or --limit=Y. Flush immediately in a background worker.
+        $this->client->flush();
```
Jean85: I would move this `flush` to a separate event (probably `WorkerMessageHandledEvent`?) so that it flushes anyway; in this way, it will flush even events fired manually by the user, or notices/deprecations, even without errors.

composer.json
emarref: Ah, I just removed the dependency. But I suppose it does need to be there. I'm not sure what the best approach is here. On the one hand, we need to ensure the correct version of Messenger, as versions <4.3 don't fire the events we need. But Messenger is not compatible with 3.4. Any ideas?

src/EventListener/MessengerListener.php
emarref: Making this optional is a good idea. I personally would have it tracked once to avoid noise.

src/EventListener/MessengerListener.php
emarref: 👍  Good point

emarref: Ah, `WorkerMessageHandledEvent` is never dispatched for failed messages.  I'll re-add the flush to the failed event, and also keep it in the handled event.

composer.json
Jean85: I would leave the package as `require-dev`, and skip it under the 3.4 CI job (there's already a regex with some exceptions). Then, if needed, we could skip failing tests under `! class_exists` against one of the component's class.

src/DependencyInjection/Configuration.php
```diff
@@ -84,6 +84,8 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->defaultValue([])
             ->prototype('scalar');
         $optionsChildNodes->scalarNode('http_proxy');
+        $optionsChildNodes->booleanNode('capture_messenger_soft_fails')
```
Jean85: I would move this option up a level: all the other options map directly to the base SDK option, this is specific to the bundle instead.

emarref:
I have modified the configuration to be under a `messenger` property, and be enabled/disabled manually. Rather than checking for the Message Bus interface.

src/DependencyInjection/Configuration.php
emarref: I have modified the configuration to be under a `messenger` property, and be enabled/disabled manually. Rather than checking for the Message Bus interface automagically. It'll default to true/false depending on whether the message bus interface is available.

composer.json
emarref: I've tidied this up using the suggested regex.

There is now a more difficult conflict to resolve - [1001.5](https://travis-ci.org/github/getsentry/sentry-symfony/jobs/664605365) shows the install failing due to a conflict with event dispatcher 3.4, and PHP 7.1.

What is your preferred solution to this?

Jean85: Ouch, so it's completely uninstallable under 3.4. Well, then we need to remove it from that job, probably with a `if SYMFONY_VERSION == 3.4 then compose remove --dev symfony/messenger` in the `before_install` step.

That would be helpful to also check that the bundle works without it too.

src/DependencyInjection/SentryExtension.php
```diff
@@ -160,6 +163,19 @@ private function configureErrorListener(ContainerBuilder $container, array $proc
...
+    {
+        if (! $processedConfiguration['enabled']) {
+            $container->removeDefinition(MessengerListener::class);
+
+            return;
+        }
+
+        if (! $processedConfiguration['capture_soft_fails']) {
+            $container->getDefinition(MessengerListener::class)->setArgument(1, false);
+        }
```
Jean85:
```suggestion
        $container->getDefinition(MessengerListener::class)->setArgument(1, $processedConfiguration['capture_soft_fails']);
```

rogamoore:
Thanks for this PR! :)
Is there an estimate when this will be released?

Jean85:
I'm waiting for #322 to be merged before preparing the 3.5 release.

src/EventListener/MessengerListener.php
```diff
@@ -0,0 +1,62 @@
...
+        if (! $this->captureSoftFails && $event->willRetry()) {
+            // Don't capture soft fails. I.e. those that will be scheduled for retry.
+            return;
+        }
+
+        $error = $event->getThrowable();
+
+        if ($error instanceof HandlerFailedException && null !== $error->getPrevious()) {
+            // Unwrap the messenger exception to get the original error
+            $error = $error->getPrevious();
```
B-Galati, line 46: `HandlerFailedException` handles a collection of exceptions so getting only the previous may lose some information along the way.

Jean85: What's the use case there? What's the situation where the handler will report multiple exceptions?

emarref: https://github.com/symfony/messenger/blob/master/Middleware/HandleMessageMiddleware.php#L56-L69

Looks like when a command has multiple handlers, they will all run. Any throwable is caught and the next handler is continued.

So here, we should iterate over `$error->getNestedExceptions()` and capture them all.

B-Galati: Or just capture `HandlerFailedException`? Not sure about the pros and cons.

Jean85: Since `HandlerFailedException` is not naturally comprehensible to Sentry, capturing it isn't ideal, it shouldn't work out of the box: the event will not show enough information on the specific handlers nor the specific errors.

Just to be sure that I understood this correctly: multiple handlers will run together only if they are all sync and not async, right? If that's the case, at least it should be a non-so-common use case, so we don't have to rush the patch.

Opening #338 to track this issue.

B-Galati: >Just to be sure that I understood this correctly: multiple handlers will run together only if they are all sync and not async, right?

No, I don't think so by reading the code and from what I understand about Messenger. Sync or Async does not matter. The only thing that matter is there are multiple handlers for a Message or not.

emarref: > Sync or Async does not matter. The only thing that matter is if there are multiple handlers for a Message or not.

Correct. Each handler for a command will be run either to completion or to an exception. Subsequent handlers will be executed and previous exceptions aggregated and added to the HandlerFailedException. So even if one handler succeeds, another handler may fail and raise the exception.

Jean85: Sorry, maybe I did not explain myself completely. For sync and async I was referring to the transport: if it's sync, the message is handled in the same request/process, and so it requires special handling on Sentry's side. If it's async, it should go through a queue (Rabbit? SQS?) so it will be executed inside a different process, and it's isolated as far as Sentry is concerned.

Does this still hold?

emarref: Yeah if a command has multiple handlers, sync or async won’t matter. The handle middleware will execute them all in the same process, catching and aggregating errors as they’re thrown, to be later raised in the handler failed exception.

Jean85:
@emarref is the `capture_soft_fails: false` working for you? I just enabled it but I'm still receiving errors...

emarref:
Yeah I've just confirmed it's working as expected for me. `symfony/symfony:4.4.8`, `sentry/sentry-symfony:3.5.1`, `sentry/sdk:2.1.0`.

Jean85:
Ok great. I discovered that the issue is in the fact that `willRetry` stays false when you use the `message:failed:retry` command, even if it will be retried later.

Jean85:
Oh crap, issue found: `willRetry` is populated in AFTER this listener runs, here: https://github.com/symfony/symfony/blob/5c056c18fc64d187c232ef0f0f0f7a2cc6da4d04/src/Symfony/Component/Messenger/EventListener/SendFailedMessageForRetryListener.php#L54-L58

```$ bin/console debug:eve Symfony\\Component\\Messenger\\Event\\WorkerMessageFailedEvent

Registered Listeners for "Symfony\Component\Messenger\Event\WorkerMessageFailedEvent" Event
===========================================================================================

 ------- ---------------------------------------------------------------------------------------------------------- ---------- 
  Order   Callable                                                                                                   Priority  
 ------- ---------------------------------------------------------------------------------------------------------- ---------- 
  #1      Sentry\SentryBundle\EventListener\MessengerListener::onWorkerMessageFailed()                               128       
  #2      Symfony\Component\Messenger\EventListener\SendFailedMessageForRetryListener::onMessageFailed()             100       
  #3      Symfony\Component\Messenger\EventListener\SendFailedMessageToFailureTransportListener::onMessageFailed()   -100      
 ------- ---------------------------------------------------------------------------------------------------------- ---------- 
```

@emarref what do you get if you call `bin/console debug:eve Symfony\\Component\\Messenger\\Event\\WorkerMessageFailedEvent`? Did you customize the priority?

Jean85:
I've opened #353 to dig further, it seems something changed between 4.3 and 4.4.

emarref:
> Did you customize the priority?

No I haven't.

> what do you get if you call bin/console debug:eve Symfony\\Component\\Messenger\\Event\\WorkerMessageFailedEvent?

```
"Symfony\Component\Messenger\Event\WorkerMessageFailedEvent" event
------------------------------------------------------------------

 ------- ---------------------------------------------------------------------------------------------------------- ---------- 
  Order   Callable                                                                                                   Priority  
 ------- ---------------------------------------------------------------------------------------------------------- ---------- 
  #1      Sentry\SentryBundle\EventListener\MessengerListener::onWorkerMessageFailed()                               128       
  #2      Symfony\Component\Messenger\EventListener\SendFailedMessageForRetryListener::onMessageFailed()             100       
  #3      Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber::onWorkerMessageFailed()      0         
  #4      Symfony\Component\Messenger\EventListener\SendFailedMessageToFailureTransportListener::onMessageFailed()   -100      
 ------- ---------------------------------------------------------------------------------------------------------- ---------- 
```

I am definitely seeing the error thrown in my handler appearing multiple times in Sentry for a single run. But the listeners don't seem to be in the right order as you've shown. I will take a look when I get more time over the next few days.

Thanks for opening the issue.

PR #329, opened 2020-04-06
Add config example for class_serializers

alexgt9:
It takes some time for me to guess exactly how to configure class_serializers, errors shown by the compiler aren't beautiful in this case.

Maybe can be useful to add this simple doc for that powerful option.

What do you think?


alexgt9:
> I like this change 👍 I would also add a code example with a class serializer excerpt, so that the explanation is clearer.

Sure! I just added it @Jean85

PR #339, opened 2020-05-06
Allow use of current hub scope in Messenger Listener

sh41:
Hopefully this is the right way to do things. Thanks to @emarref for the Messenger listener - I had no idea how much I was missing from these exceptions being swallowed. I would like to add another listener with a higher priority that adds user details to the scope, but found that there was no way to access the scope of the Client in the new listener. I've changed the listener so that it uses the currenthub. This allows me to create a higher priority listener that will populate the scope with something like the below.

```php
<?php

namespace App\EventListener;

use App\Repository\UserRepository;
use Sentry\SentrySdk;
use Sentry\State\Scope;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\Messenger\Event\WorkerMessageFailedEvent;

class SentryMessengerListener implements EventSubscriberInterface
{
    /**
     * @var bool
     */
    private $captureSoftFails;
    /**
     * @var UserRepository
     */
    private UserRepository $userRepository;

    /**
     * @param UserRepository $userRepository
     * @param bool $captureSoftFails
     */
    public function __construct(UserRepository $userRepository, bool $captureSoftFails = true)
    {
        $this->userRepository = $userRepository;
        $this->captureSoftFails = $captureSoftFails;
    }

    /**
     * @param WorkerMessageFailedEvent $event
     */
    public function onWorkerMessageFailed(WorkerMessageFailedEvent $event): void
    {
        if (! $this->captureSoftFails && $event->willRetry()) {
            // Don't capture soft fails. I.e. those that will be scheduled for retry.
            return;
        }
        // If hte message has a userId, look up the user and add it to the current hub's scope.
        $message = $event->getEnvelope()->getMessage();
        $userId = $message->userId??null;
        if (!$userId) {
            return;
        }
        $user = $this->userRepository->find($userId);
        if ($user) {
            $userData = [
                'username' => $user->getUsername(),
            ];
            SentrySdk::getCurrentHub()
                ->configureScope(
                    function (Scope $scope) use ($userData): void {
                        $scope->setUser($userData, true);
                    }
                );
        }
    }

    /**
     * @return array
     */
    public static function getSubscribedEvents()
    {
        return [
            WorkerMessageFailedEvent::class => [ 'onWorkerMessageFailed', 200 ],
        ];
    }
}
```

I think I've got all of the tests working. Open to feedback, or another way of accessing scope for the listeners. Please just let me know.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -271,16 +271,6 @@ parameters:
 			path: test/EventListener/ConsoleListenerTest.php
 
 		-
-			message: "#^Property Sentry\\\\SentryBundle\\\\Test\\\\EventListener\\\\MessengerListenerTest\\:\\:\\$client has no typehint specified\\.$#"
-			count: 1
-			path: test/EventListener/MessengerListenerTest.php
-
-		-
-			message: "#^Property Sentry\\\\SentryBundle\\\\Test\\\\EventListener\\\\MessengerListenerTest\\:\\:\\$hub has no typehint specified\\.$#"
-			count: 1
-			path: test/EventListener/MessengerListenerTest.php
-
-		-
 			message: "#^Class Symfony\\\\Component\\\\Messenger\\\\Event\\\\WorkerMessageHandledEvent constructor invoked with 3 parameters, 2 required\\.$#"
 			count: 1
 			path: test/EventListener/MessengerListenerTest.php
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -276,7 +276,7 @@ parameters:
 			path: test/EventListener/MessengerListenerTest.php
 
 		-
-			message: "#^Property Sentry\\\\SentryBundle\\\\Test\\\\EventListener\\\\MessengerListenerTest\\:\\:\\$currentHub has no typehint specified\\.$#"
+			message: "#^Property Sentry\\\\SentryBundle\\\\Test\\\\EventListener\\\\MessengerListenerTest\\:\\:\\$hub has no typehint specified\\.$#"
 			count: 1
 			path: test/EventListener/MessengerListenerTest.php
 
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -276,6 +276,11 @@ parameters:
 			path: test/EventListener/MessengerListenerTest.php
 
 		-
+			message: "#^Property Sentry\\\\SentryBundle\\\\Test\\\\EventListener\\\\MessengerListenerTest\\:\\:\\$currentHub has no typehint specified\\.$#"
+			count: 1
+			path: test/EventListener/MessengerListenerTest.php
+
+		-
 			message: "#^Class Symfony\\\\Component\\\\Messenger\\\\Event\\\\WorkerMessageHandledEvent constructor invoked with 3 parameters, 2 required\\.$#"
 			count: 1
 			path: test/EventListener/MessengerListenerTest.php
--- a/src/DependencyInjection/SentryExtension.php
+++ b/src/DependencyInjection/SentryExtension.php
@@ -174,7 +174,7 @@ class SentryExtension extends Extension
             return;
         }
 
-        $container->getDefinition(MessengerListener::class)->setArgument(1, $processedConfiguration['capture_soft_fails']);
+        $container->getDefinition(MessengerListener::class)->setArgument(0, $processedConfiguration['capture_soft_fails']);
     }
 
     /**
--- a/src/DependencyInjection/SentryExtension.php
+++ b/src/DependencyInjection/SentryExtension.php
@@ -174,7 +174,7 @@ class SentryExtension extends Extension
             return;
         }
 
-        $container->getDefinition(MessengerListener::class)->setArgument(0, $processedConfiguration['capture_soft_fails']);
+        $container->getDefinition(MessengerListener::class)->setArgument(1, $processedConfiguration['capture_soft_fails']);
     }
 
     /**
--- a/src/EventListener/MessengerListener.php
+++ b/src/EventListener/MessengerListener.php
@@ -2,7 +2,7 @@
 
 namespace Sentry\SentryBundle\EventListener;
 
-use Sentry\FlushableClientInterface;
+use Sentry\SentrySdk;
 use Symfony\Component\Messenger\Event\WorkerMessageFailedEvent;
 use Symfony\Component\Messenger\Event\WorkerMessageHandledEvent;
 use Symfony\Component\Messenger\Exception\HandlerFailedException;
@@ -10,22 +10,15 @@ use Symfony\Component\Messenger\Exception\HandlerFailedException;
 final class MessengerListener
 {
     /**
-     * @var FlushableClientInterface
-     */
-    private $client;
-
-    /**
      * @var bool
      */
     private $captureSoftFails;
 
     /**
-     * @param FlushableClientInterface $client
-     * @param bool                     $captureSoftFails
+     * @param bool $captureSoftFails
      */
-    public function __construct(FlushableClientInterface $client, bool $captureSoftFails = true)
+    public function __construct(bool $captureSoftFails = true)
     {
-        $this->client = $client;
         $this->captureSoftFails = $captureSoftFails;
     }
 
@@ -46,8 +39,11 @@ final class MessengerListener
             $error = $error->getPrevious();
         }
 
-        $this->client->captureException($error);
-        $this->client->flush();
+        $hub = SentrySdk::getCurrentHub();
+        $hub->captureException($error);
+        if (method_exists($hub->getClient(), 'flush')) {
+            $hub->getClient()->flush();
+        }
     }
 
     /**
@@ -57,6 +53,9 @@ final class MessengerListener
     {
         // Flush normally happens at shutdown... which only happens in the worker if it is run with a lifecycle limit
         // such as --time=X or --limit=Y. Flush immediately in a background worker.
-        $this->client->flush();
+        $hub = SentrySdk::getCurrentHub();
+        if (method_exists($hub->getClient(), 'flush')) {
+            $hub->getClient()->flush();
+        }
     }
 }
--- a/src/EventListener/MessengerListener.php
+++ b/src/EventListener/MessengerListener.php
@@ -42,16 +42,15 @@ final class MessengerListener
 
         $error = $event->getThrowable();
 
-        if ($error instanceof HandlerFailedException && null !== $error->getPrevious()) {
-            // Unwrap the messenger exception to get the original error
-            $error = $error->getPrevious();
+        if ($error instanceof HandlerFailedException) {
+            foreach ($error->getNestedExceptions() as $nestedException) {
+                $this->hub->captureException($nestedException);
+            }
+        } else {
+            $this->hub->captureException($error);
         }
 
-        $this->hub->captureException($error);
-        $client = $this->hub->getClient();
-        if ($client instanceof FlushableClientInterface) {
-            $client->flush();
-        }
+        $this->flush();
     }
 
     /**
@@ -61,6 +60,11 @@ final class MessengerListener
     {
         // Flush normally happens at shutdown... which only happens in the worker if it is run with a lifecycle limit
         // such as --time=X or --limit=Y. Flush immediately in a background worker.
+        $this->flush();
+    }
+
+    private function flush(): void
+    {
         $client = $this->hub->getClient();
         if ($client instanceof FlushableClientInterface) {
             $client->flush();
--- a/src/EventListener/MessengerListener.php
+++ b/src/EventListener/MessengerListener.php
@@ -2,7 +2,8 @@
 
 namespace Sentry\SentryBundle\EventListener;
 
-use Sentry\SentrySdk;
+use Sentry\FlushableClientInterface;
+use Sentry\State\HubInterface;
 use Symfony\Component\Messenger\Event\WorkerMessageFailedEvent;
 use Symfony\Component\Messenger\Event\WorkerMessageHandledEvent;
 use Symfony\Component\Messenger\Exception\HandlerFailedException;
@@ -10,15 +11,22 @@ use Symfony\Component\Messenger\Exception\HandlerFailedException;
 final class MessengerListener
 {
     /**
+     * @var HubInterface
+     */
+    private $hub;
+
+    /**
      * @var bool
      */
     private $captureSoftFails;
 
     /**
-     * @param bool $captureSoftFails
+     * @param HubInterface $hub
+     * @param bool         $captureSoftFails
      */
-    public function __construct(bool $captureSoftFails = true)
+    public function __construct(HubInterface $hub, bool $captureSoftFails = true)
     {
+        $this->hub = $hub;
         $this->captureSoftFails = $captureSoftFails;
     }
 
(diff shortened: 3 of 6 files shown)


src/EventListener/MessengerListener.php
```diff
@@ -2,30 +2,23 @@
...
      * @var bool
      */
     private $captureSoftFails;
 
     /**
-     * @param FlushableClientInterface $client
-     * @param bool                     $captureSoftFails
+     * @param bool $captureSoftFails
      */
-    public function __construct(FlushableClientInterface $client, bool $captureSoftFails = true)
```
emarref, line 26: Could you inject the `Sentry\State\HubInterface` instead? This avoids calling static/singletons later on.

src/EventListener/MessengerListener.php
```diff
@@ -46,8 +39,11 @@ public function onWorkerMessageFailed(WorkerMessageFailedEvent $event): void
             $error = $error->getPrevious();
         }
 
-        $this->client->captureException($error);
-        $this->client->flush();
+        $hub = SentrySdk::getCurrentHub();
+        $hub->captureException($error);
+        if (method_exists($hub->getClient(), 'flush')) {
```
emarref, line 44: Could you acquire the client `$client = $hub->getClient()`, then test if `$client instanceof FlushableClientInterface`, rather than the call to `method_exists`?

emarref:
For your local high-priority listener, why not configure the scope as soon as you know the user, rather than only when capturing exceptions? This would also help when pushing information to Sentry that is not an exception. E.g. breadcrumbs, or manual error captures.

Perhaps this is personal preference coming through, but I would be inclined to "select" the user somehow that makes sense to your app. E.g. in some service, which fires an event alerting the app. You'd then have a listener on that "user selected" event that configures the Sentry scope.

phpstan-baseline.neon
```diff
@@ -275,6 +275,11 @@ parameters:
 			count: 1
 			path: test/EventListener/MessengerListenerTest.php
 
+		-
+			message: "#^Property Sentry\\\\SentryBundle\\\\Test\\\\EventListener\\\\MessengerListenerTest\\:\\:\\$currentHub has no typehint specified\\.$#"
```
Jean85, line 279: :-1: on ignoring static analysis errors.

src/DependencyInjection/SentryExtension.php
```diff
@@ -174,7 +174,7 @@ private function configureMessengerListener(ContainerBuilder $container, array $
             return;
         }
 
-        $container->getDefinition(MessengerListener::class)->setArgument(1, $processedConfiguration['capture_soft_fails']);
+        $container->getDefinition(MessengerListener::class)->setArgument(0, $processedConfiguration['capture_soft_fails']);
```
Jean85, line 177: I would be against this change. This is needed to run the listener before any default one, but if you need to add info of your own before, you can just register a listener with an higher priority of 2 or more.

Also, I agree with @emarref suggestion: if you need to attach user information, you just need to do that as soon as the user is known and push it into the Scope; then, IF an event is fired, the info will be attached to the event.

To be complete and thorough, this bundle does that already, but you need to enable `send_default_pii`, see https://github.com/getsentry/sentry-symfony/blob/6c3f3236504c2722eb5a8b26a10a2596bee62494/src/EventListener/RequestListener.php#L54-L87

src/EventListener/MessengerListener.php
Jean85: > Could you inject the `Sentry\State\HubInterface` instead? This avoids calling static/singletons later on.

Agreed. You should copy the behavior of the other listeners here.

src/EventListener/MessengerListener.php
Jean85: > Could you acquire the client `$client = $hub->getClient()`, then test if `$client instanceof FlushableClientInterface`, rather than the call to `method_exists`?

:+1: here too.

src/EventListener/MessengerListener.php
```diff
@@ -57,6 +53,9 @@ public function onWorkerMessageHandled(WorkerMessageHandledEvent $event): void
     {
         // Flush normally happens at shutdown... which only happens in the worker if it is run with a lifecycle limit
         // such as --time=X or --limit=Y. Flush immediately in a background worker.
-        $this->client->flush();
+        $hub = SentrySdk::getCurrentHub();
+        if (method_exists($hub->getClient(), 'flush')) {
+            $hub->getClient()->flush();
+        }
```
Jean85, line 59: `instanceof FlushableClient` here too.

src/EventListener/MessengerListener.php
sh41: Hi both, I must admit that I thought I had followed the behaviour of the Console Listener, but I now see that it uses both DI and Static to get the hub. I didn't see the DI method, but I also much prefer it. I've made this change.

Jean85:
Also, please check #340.

phpstan-baseline.neon
sh41: I've never used phpstan before, and confess that I was just copying the way that the rest of that file deals with the errors it has. 

The specific error here is that there is no typehint, which I think only became a thing in PHP 7.4? I can make the change, but I think that will stop us from being able to run tests on <7.4?

src/DependencyInjection/SentryExtension.php
sh41: The change here was because I had removed the `FlushableClientInterface` argument to the constructor, so the compiler pass needed to insert the configuration option in argument 0 rather than argument 1. Now that we are injecting the Hub the configuration argument has moved back to it's original position, so this has been reverted.

src/EventListener/MessengerListener.php
sh41: Yes this is now done.

phpstan-baseline.neon
Jean85: PHPStan is also fine if you specify the type with a `@var` PHPDoc.

sh41: Thanks, I've reverted the change to phpstan ignore file and typehinted both the hub and the client property so that they are consistent with each other.

sh41:
> Also, please check #340.

@Jean85 What's the best approach with #340? Should I wait until it has been approved and then rework this PR around it? It is all very much in the same area.

sh41:
> For your local high-priority listener, why not configure the scope as soon as you know the user, rather than only when capturing exceptions? This would also help when pushing information to Sentry that is not an exception. E.g. breadcrumbs, or manual error captures.
> 
> Perhaps this is personal preference coming through, but I would be inclined to "select" the user somehow that makes sense to your app. E.g. in some service, which fires an event alerting the app. You'd then have a listener on that "user selected" event that configures the Sentry scope.

I had been thinking of adding to the scope using Middleware rather than an event listener. My consumers all run asyncronously on different VMs to the message producers, so the only information for scope that I have comes from the message itself. There is probably a clever way to add information via stamps and then use those stamps in some middleware to populate scope for a particular message.

emarref:
Yeah sounds like middleware is the way to go for your use case. Just remember that your middleware should unset what it adds to the scope after the command has been handled or if it fails.

src/EventListener/MessengerListener.php
sh41: Yes done.

sh41: done

Jean85:
> Yeah sounds like middleware is the way to go for your use case. Just remember that your middleware should unset what it adds to the scope after the command has been handled or if it fails.

Unset is overkill, you could just `Scope::push` at the start and `Scope::pop` at the end, and you're good to go. Push adds a new scope to the stack with a copy of all the preexisting data, pop removes that scope from the stack, basically doing a full undo.

> @Jean85 What's the best approach with #340? Should I wait until it has been approved and then rework this PR around it? It is all very much in the same area.

Due to my response above :arrow_up: I think that you should wait for #340; with that plus those advices you'll probably be able to do everything you need. Ping me if you need more suggestions.

sh41:
I've merged in the changes from #340. I believe that this PR is still valuable because without it the hub's scope is not included in any exception reported back to Sentry. By changing the listener to use [`HubInterface::captureException`](https://github.com/getsentry/sentry-php/blob/97bbca2c828651dfbd04769c94c1cc8b37e096a3/src/State/Hub.php#L132) rather than [`FlushableClientInterface::captureException` ](https://github.com/getsentry/sentry-php/blob/97bbca2c828651dfbd04769c94c1cc8b37e096a3/src/Client.php#L94) any scope that has been added to the hub will be sent with the flush. I can't see any other way of adding scope, but happy to be steered right if I am mistaken with that?

sh41:
Thanks @Jean85 & thanks again to @emarref for the initial PR. I have found so many bugs that we just couldn't see before. 

For the benefit of anyone looking to add UserContext to their Messenger exceptions, I have ended up with middleware to configure the scope. It uses a stamp to keep the user details with the envelope. If there is currently a logged in user it will stamp the envelope appropriately. Alternatively in a situation where I don't have logged in user I can use something like this: 

```php
$bus->dispatch(new Message(), [new UserContextStamp($userId, $username)]);
```
These are my Stamp and Middleware classes. Feel free to use. YMMV. 
```php
namespace App\Messenger\Stamp;

use Symfony\Component\Messenger\Stamp\StampInterface;

class UserContextStamp implements StampInterface
{
    public int $id;
    public string $username;

    public function __construct(int $id, string $username)
    {
        $this->id = $id;
        $this->username = $username;
    }
}
```

```php
namespace App\Messenger\Middleware;

use App\Messenger\Stamp\UserContextStamp;
use Sentry\State\HubInterface;
use Symfony\Component\Messenger\Envelope;
use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
use Symfony\Component\Messenger\Middleware\StackInterface;
use Symfony\Component\Messenger\Stamp\HandledStamp;
use Symfony\Component\Security\Core\Security;

final class SentryUserContextMiddleware implements MiddlewareInterface
{
    /**
     * @var HubInterface
     */
    private HubInterface $hub;
    /**
     * @var Security
     */
    private Security $security;

    /**
     * @param HubInterface $hub
     * @param Security $security
     */
    public function __construct(HubInterface $hub, Security $security)
    {
        $this->hub = $hub;
        $this->security = $security;
    }

    public function handle(Envelope $envelope, StackInterface $stack): Envelope
    {
        // If we haven't previously had a UserContextStamp added and Security is
        // able to provide a suitable user, add it to the envelope.
        if (null === $envelope->last(UserContextStamp::class)) {
            $user = $this->security->getUser();
            if ($user instanceof UserInterface) {
                $envelope = $envelope->with(new UserContextStamp($user->getId(), $user->getUsername()));
            }
        }

        // Push a new scope and then configure it with the user context if we have one.
        $scope = $this->hub->pushScope();

        // The stamp might have just been added above, or might have been added during dispatch,
        // in other middleware or any other time before now.
        /** @var $usernameStamp UserContextStamp */
        if (null !== ($usernameStamp = $envelope->last(UserContextStamp::class))) {
            $userContext['id'] = $usernameStamp->id;
            $userContext['username'] = $usernameStamp->username;
            $scope->setUser($userContext, true);
        }
        try {
            $envelope = $stack->next()->handle($envelope, $stack);
            return $envelope;
        } finally {
            // if we successfully handled the message we can clear the scope
            if (null !== $envelope->last(HandledStamp::class)) {
                // pop scope so that the hub is all reset ready for the next message coming through.
                $this->hub->popScope();
            }
        }
    }
}
```

There is a slight inefficiency with this, which I'm not sure how to solve, which is that the new MessengerListener only triggers after the entire middleware stack has finished executing. That means that if I want my scope to still be in the hub's stack when a WorkerMessageFailed event is fired I can't pop it. That's what the finally part of the middleware above is for. If the message was properly handled we can lose the scope, but if it is not properly handled we need to keep it for the MessengerListener call which comes afterwards. I think I might need to add a lower priority listener for WorkerMessageFailed to pop the scope.

emarref:
> the new MessengerListener only triggers after the entire middleware stack has finished executing

True.

However, the danger with this approach is that your scope will never be popped for failed messages. Subsequent messages failing on the same worker will get deeper and deeper in scopes. Unless you have a custom listener on the handler failed event to pop the scope. But this seems disconnected and messy.

Perhaps a better approach instead of middleware is a custom subscriber in your app that listens to the `WorkerMessageReceivedEvent` event to grab the UserContextStamp and push the scope and configure it with user information. You can then pop the scope by listening for the `WorkerMessageFailedEvent` and `WorkerMessageHandledEvent` events.

In fact, it might even be something that could be added to the bundle.

Jean85:
> In fact, it might even be something that could be added to the bundle.

I agree :+1: Is it possible to push/pop around a single handler too or there is no event to listen in that case?

emarref:
> Is it possible to push/pop around a single handler too or there is no event to listen in that case?

There is no event, so I don't think there is a way to do this.

https://github.com/symfony/messenger/blob/master/Middleware/HandleMessageMiddleware.php#L56-L69

sh41:
I dropped  the Middleware in favour of an event based system. The below is what I have right now. The `SendMessageToTransportsEvent` is used on the producer side to extract user information from the security service if there isn't already a [`UserContextStamp`](https://github.com/getsentry/sentry-symfony/pull/339#issuecomment-625309775) attached.

On my async worker side of things the scope is populated from the details in the stamp as soon as the message is received and then the scope is popped when the message processing completes, either by a successful handling or by a failure. 

```php
namespace App\EventListener;

use App\Messenger\Stamp\UserContextStamp;
use Sentry\State\HubInterface;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\Messenger\Event\SendMessageToTransportsEvent;
use Symfony\Component\Messenger\Event\WorkerMessageFailedEvent;
use Symfony\Component\Messenger\Event\WorkerMessageHandledEvent;
use Symfony\Component\Messenger\Event\WorkerMessageReceivedEvent;
use Symfony\Component\Security\Core\Security;

class AddSentryContextToMessengerListener implements EventSubscriberInterface
{
    /**
     * @var HubInterface
     */
    private HubInterface $hub;
    /**
     * @var Security
     */
    private Security $security;

    /**
     * @param HubInterface $hub
     * @param Security $security
     */
    public function __construct(HubInterface $hub, Security $security)
    {
        $this->hub      = $hub;
        $this->security = $security;
    }

    public function onSendMessageToTransports(SendMessageToTransportsEvent $event)
    {
        $envelope = $event->getEnvelope();
        // If we haven't previously had a UserContextStamp added and Security is
        // able to provide a suitable user, add it to the envelope.
        if (null === $envelope->last(UserContextStamp::class)) {
            $user = $this->security->getUser();
            if ($user instanceof UserInterface) {
                $envelope = $envelope->with(new UserContextStamp($user->getId(), $user->getUsername()));
            }
        }

        $event->setEnvelope($envelope);
    }

    public function onWorkerMessageReceived(WorkerMessageReceivedEvent $event)
    {
        $envelope = $event->getEnvelope();
        // Push a new scope and then configure it with the user context if we have one.
        // we alway push a new scope, because we always pop the scope on completion.
        $scope = $this->hub->pushScope();

        // The stamp might have just been added above, or might have been added during dispatch,
        // or any other time before now.
        /** @var $usernameStamp UserContextStamp */
        if (null !== ($usernameStamp = $envelope->last(UserContextStamp::class))) {
            $userContext['id']       = $usernameStamp->id;
            $userContext['username'] = $usernameStamp->username;
            $scope->setUser($userContext, true);
        }
    }

    public function onWorkerMessageFailedOrHandled(): void
    {
        // runs after the built in listener to clear the scope.
        $this->hub->popScope(); // clear the scope from the SentryUserContextMiddleware here.
    }

    /**
     * @return array
     */
    public static function getSubscribedEvents()
    {
        return [
            SendMessageToTransportsEvent::class => ['onSendMessageToTransports'],
            WorkerMessageReceivedEvent::class   => ['onWorkerMessageReceived'],
            WorkerMessageFailedEvent::class     => ['onWorkerMessageFailedOrHandled', 127],
            WorkerMessageHandledEvent::class    => ['onWorkerMessageFailedOrHandled', 127],
        ];
    }
}
```

emarref:
Looks good

Issue #338, opened 2020-05-06
HandlerFailedException could wrap multiple exceptions


Jean85:
As pointed out by @B-Galati and following comments [here](https://github.com/getsentry/sentry-symfony/pull/326#discussion_r419865622), the Symfony Messenger could have multiple synchronous handlers for the same message. If multiple handlers fail in this case, the resulting single `HandlerFailedException` would wrap all the raised exceptions.

To fix this, we probably have to iterate over `$error->getNestedExceptions()` and capture them all.

PR #340, opened 2020-05-07
Capture all messenger exceptions

emarref:
When a command has multiple handlers, if any of those handlers throw an error, other handlers will continue. Multiple handlers may throw errors. This PR ensures all thrown errors are captured.

Addresses https://github.com/getsentry/sentry-symfony/issues/338

### DIFF
--- a/src/EventListener/MessengerListener.php
+++ b/src/EventListener/MessengerListener.php
@@ -41,14 +41,12 @@ final class MessengerListener
 
         $error = $event->getThrowable();
 
-        if (! $error instanceof HandlerFailedException) {
-            // All errors thrown during handling a command are captured by the HandleMessageMiddleware,
-            // and are raised inside a HandlerFailedException. Type check for safety.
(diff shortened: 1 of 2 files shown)


src/EventListener/MessengerListener.php
```diff
@@ -57,6 +59,11 @@ public function onWorkerMessageHandled(WorkerMessageHandledEvent $event): void
     {
         // Flush normally happens at shutdown... which only happens in the worker if it is run with a lifecycle limit
         // such as --time=X or --limit=Y. Flush immediately in a background worker.
+        $this->flush();
+    }
+
+    private function flush(): void
+    {
         $this->client->flush();
```
Jean85, line 67: Shouldn't we check `if $this->client instanceof FlushableClient`?

emarref: The constructor only accepts a flushable client

Issue #346, opened 2020-05-25
Breaking versions with composer/package-versions-deprecated


NikitaKharkov:
So, I can't update this package to 3.5 because of using this package. This package is installed automatically to version 1.8 (current and the single one), so can't remove it or do something with it. At the same, time your package has the dependency:
```
    "conflict": {
        "ocramius/package-versions": "<1.3"
    },
```
Can you fix it? Or can you advise me on what to do?

Jean85:
First thing, you should try and check if you're using the latest Composer patch, 1.10.6, which contains a fix for resolving the dependecies of `jean85/pretty-package-versions`, which in turn requires `ocramius/package-versions`. 

You should be able to install 3.5 without issues then. But, nonetheless, we should drop that requirement somehow, because it's better to use the native Composer 2 support in the future.

jakub-frajt:
Even with latest version of Composer (1.10.7) it's not possible to install this package.

Jean85:
Can you share a `composer.json` that reproduces this issue? I cannot understand why you can't, the constraints shouldn't create any issue.

Jean85:
As pointed out by #347, there was an underlying issue, but it has been fixed with https://github.com/composer/package-versions-deprecated/pull/3.

Closing as fixed, feel free to reply or reopen if it still doesn't work.

PR #347, opened 2020-06-18
Remove "conflict" section from composer.json

nicolas-grekas:
This line is not required because the conflicting package is already in the "require" section.

Moreover, this line creates issue #346 

In practice, this line makes this package conflict with `jean85/pretty-package-versions` v1.3.*, which means it's not possible to generate a composer.lock file that is compatible with both composer 1 and composer 2.

:bomb: :boom:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,9 +48,6 @@
         "symfony/process": "^3.4||^4.0||^5.0",
         "symfony/yaml": "^3.4||^4.0||^5.0"
     },
-    "conflict": {
-        "ocramius/package-versions": "<1.3"
-    },
     "suggest": {
         "monolog/monolog": "Required to use the Monolog handler"
     },


nicolas-grekas:
Oh, note that this targets master which is aliased to v4, but the patch should be released as v3.5.3 really.

Thanks!

nicolas-grekas:
See also https://github.com/composer/package-versions-deprecated/pull/3

This line effectively prevents using version 1.3 with sentry-symfony v3.5.

I don't get how you can be against the change, the deps are just broken right now.

nicolas-grekas:
Since 6c3f323, you've added `ocramius/package-versions` in the `require` section, which makes the conflict line a duplicate as far as declarations are concerned. Unless you want to explicitly forbid installing `composer/package-versions-deprecated`, this line has to be removed.

Jean85:
> I don't get how you can be against the change, the deps are just broken right now.

They are not, it's a combination of Composer inconsistency + composer/package-versions-deprecated#3

```
$ composer why-not jean85/pretty-package-versions:1.3 
There is no installed package depending on "jean85/pretty-package-versions" in versions not matching 1.3
```

But I understand what you want to fix. As I said before, the only roadblock for this change is the direct usage of the root package constant from Ocramius' package. I would need to add some kind of getter in my package to mask the differences between Ocramius', Composer's fork and Composer 2 behavior, which I already had in mind  (see https://github.com/Jean85/pretty-package-versions/issues/19).

nicolas-grekas:
Let me put it another way:
```json
{
    "require": {
        "sentry/sentry-symfony": "^3.5",
        "jean85/pretty-package-versions": "~1.3.0"
    }
}
```

This `composer.json` is not installable, although its the only potential way to generate a composer.lock file that works on Composer 1&2 while using sentry-symfony 3.5.

This is just plain broken, I hope you get it better now.

nicolas-grekas:
This
image

means the conflict is a leftover, with a nasty side-effect I just described.

Jean85:
The problem at hand seems to have been fixed in https://github.com/composer/package-versions-deprecated/pull/3 with the 1.8.1 release.

As for moving forward, I think I'll leverage https://github.com/Jean85/pretty-package-versions/pull/22 + https://github.com/Jean85/pretty-package-versions/issues/19 to bump to `jean85/pretty-package-versions": "^1.5 || ^2.0"` and be fine with it.

Jean85:
Closing in favor of tracking the issue in #349

PR #348, opened 2020-06-18
removed dev files from published package

vladyslavstartsev:
Those files and folders are never used by end client of the lib


Jean85:
Build seems to be failing for unrelated causes. I'll merge anyway and fix it later.

PR #352, opened 2020-06-28
Add full command to extras

### LINKED ISSUE #351: Capture command arguments
CLI Commands are usually run with certain arguments. Error is also most of the time being triggered with certain CLI arguments only, this is why it's important to have it. Would you welcome contribution to log not only command name, but command arguments in your ConsoleListener?

ostrolucky:
Fixes #351

### DIFF
--- a/tests/End2End/End2EndTest.php
+++ b/tests/End2End/End2EndTest.php
@@ -10,6 +10,8 @@ use Symfony\Bundle\FrameworkBundle\Client;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Bundle\FrameworkBundle\KernelBrowser;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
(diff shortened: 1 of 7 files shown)


Jean85:
@ste93cry do you think we should recover this into the 4.0 release? Take a look at the discussion in #351 too.

ste93cry:
I'm generally favourable to adding logging of the command. Regarding the sanitization, As it's a full string I don't think that Sentry can hide or detect specific portions of it so that values of certain command arguments or options will be hidden... I wonder anyway why someone would pass sensitive data as an argument or option, since it would be visible in the CLI as much as in any log system. Bonus question: I'm really doubtful about whether the command should be logged in the extra context or as a breadcrumb

ostrolucky:
According link in linked issue, laravel also passes whole thing as string, so I don't believe this is an issue. 

As for breadcrumbs, in my pov breadcrumbs are something which happens in sequence, like a log. I think command name does not quite fit in that definition, similarly like request url and so on. Especially when command name already is a tag instead of bread crumb and I'm implementing now just logging of full command name - but I deemed full command name being too unique for being a tag, hence second best was extra.

ste93cry:
>As for breadcrumbs, in my pov breadcrumbs are something which happens in sequence, like a log. I think command name does not quite fit in that definition, similarly like request url and so on

True, but in Symfony the concept of subrequest exists as it exists the concept of a command started and running within another. I would say that logging as bredcrumbs both the start of a request and of a command would be useful. Anyway, I agree that at the moment logging the full command in the extras is enough

>I deemed full command name being too unique for being a tag

Absolutely agree with this claim

tests/End2End/End2EndTest.php
```diff
@@ -171,6 +173,25 @@ public function testNotice(): void
         $this->assertEventCount(1);
     }
 
+    public function testCommand(): void
```
ste93cry, line 176: Rather than an integration test I would expect the unit tests of the `ConsoleCommandListener` class to be updated

ste93cry (changes requested):

tests/End2End/End2EndTest.php
ostrolucky: Unit test wouldn't test if command name was captured. For example, there is no guarantee input type that is received in listener won't be changed in future to something else than ArgvInput. But in unit test we would have to hardcode ArgvInput mock, so change won't be picked up. End2end test is more useful here; also as future basis, as currently there are no end2end command tests. Would be shame to throw that work away.

Jean85 (changes requested):
LGTM :+1: but please add a changelog entry

tests/End2End/End2EndTest.php
ste93cry: There is no relation between the presence of the command name in the logged string and the code of the listener in my opinion: the only thing that it does is casting the input object to a `string`. I don't say that this e2e should be removed, but I would like to add a unit test too for coherence, even if now they overlaps

ostrolucky: Ok, unit test added

Jean85:
Build should be fixed by #480, but there are conflicts now.

ostrolucky:
rebased

ste93cry (approved):

Jean85 (approved):

PR #353, opened 2020-07-03
Messenger: capture_soft_fails:false doesn't work

Jean85:
This stems from https://github.com/getsentry/sentry-symfony/pull/326#issuecomment-653455458: I've noticed that the `capture_soft_fails` option is not working for me, because the `willRetry` flag is set later, in a different listener with a priority of 100, where our listener has a default priority of 128.


Jean85:
Ok, the bug seems to be absent on 4.3: https://travis-ci.org/github/getsentry/sentry-symfony/jobs/704627016

Jean85:
@emarref this is ready, can you test it out?

Jean85:
I personally tested this in production, it works.

emarref:
Sorry about that. I've been under the gun with a deadline. It looks like it should work well. I'm still not sure why it was working OK for me in the first place.

PR #361, opened 2020-08-07
Make the RequestIntegration integration fetch the request from the request stack

ste93cry:
With this PR the `RequestIntegration` integration now fetches the HTTP request from the request stack of Symfony instead of creating a new request from the PHP superglobals. This should (but I didn't test if it was really broken before) also solve the fact that in the context of a Symfony sub-request the request data attached to the event was referring to the main PHP request


Jean85:
@ste93cry should we rebase and merge this?

ste93cry:
Yes, I will do it after a few other pending PRs get merged

PR #377, opened 2020-10-02
Replace Travis CI with GitHub Actions

Jean85:


ste93cry:
What's the benefit of this change? I like GitHub Actions even though I didn't use them yet, but I don't see how this could benefit us. Also, much more importantly, does the action you are going to use support Windows? I know we don't test the bundle on it and it doesn't matter so much compared to Linux builds, but it's a thing I would like to do since a long time

Jean85:
I wanted to abandon Travis since the acquisition debacle (https://news.ycombinator.com/item?id=19218036); also their UX is getting less and less responsive.

GA supports Windows, yeah, and it's a lot more fast and responsive, probably due to the tight integration.

ste93cry:
>I wanted to abandon Travis since the acquisition debacle

I would not change a tool that still works great for political reasons, and honestly I see no benefit with this change apart from the fact that it's part of GitHub itself 😉 Said this, I don't intend to block this PR, go ahead if you think it's worth it

Jean85:
Politics aside, it works better and faster: A full run in 1:36 with all checks in parallel: https://github.com/getsentry/sentry-symfony/actions/runs/285923001

stayallive:
@Jean85 I did notice the Zeus webhooks missing...

I would have first checked with @HazAT how important those are for the release cycle since I believe Craft does check in with Zeus. 

If you did check that beforehand don't mind me but otherwise this might have to be reverted.

Jean85:
@stayallive my bad, I forgot about it. Shouldn't be possible to have the same hook from GA too? @HazAT ?

HazAT:
Yes, Zeus also works with GHA

https://github.com/getsentry/sentry-javascript/blob/master/.github/workflows/build.yml#L121

stayallive:
Ah if we don't need Zeus we're good then 👍

PR #381, opened 2020-10-14
Use a FQCN for Sentry bundle in AppKernel

melkamar:
Without the leading slash the import does not work when
the AppKernel class is in a non-root namespace.


Jean85 (approved):
Thanks for spotting this!

PR #385, opened 2020-11-07
Update the dependencies and require the SDK 3.x

ste93cry:
As per title, along with a few dependencies (both dev and non) I updated the main SDK to version 3.x 🎉 There weren't many changes to do and the majority were in the tests. In reality, I just made the minimum amount of work to make the tests pass, taking for granted that since both unit and e2e tests are green everything is fine. If someone that is brave enough wants to try this out before merging it it would be really cool


ste93cry:
After hours and hours of debugging, I found out the cause of the broken build: a change that happened between PHP `7.2` and `7.3` (I didn't find which one) which made sure that when an error is caught in a `try` statement, `error_get_last()` is not filled with its information. However, Xdebug maintained this behavior until `2.9.3` when [the bug](https://bugs.xdebug.org/1758) was fixed. TLDR, Travis CI is using a bugged version of Xdebug for PHP 7.2. I have no clue about why the step with `SYMFONY_VERSION=3.4.* SYMFONY_DEPRECATIONS_HELPER=disabled` doesn't fail though

composer.json
```diff
@@ -15,14 +24,11 @@
...
-    "config": {
-        "sort-packages": true
-    },
     "require": {
         "php": "^7.1",
-        "jean85/pretty-package-versions": "^1.0",
-        "ocramius/package-versions": "^1.3.0",
-        "sentry/sdk": "^2.1",
+        "jean85/pretty-package-versions": "^1.5",
+        "ocramius/package-versions": "^1.8",
```
Jean85, line 30: IIRC, since you bumped `jean85/pretty-package-versions` to 1.5, Ocramius' package could be removed.

test/EventListener/MessengerListenerTest.php
```diff
@@ -14,7 +15,7 @@
 
 class MessengerListenerTest extends BaseTestCase
 {
-    /** @var \Prophecy\Prophecy\ObjectProphecy|FlushableClientInterface */
+    /** @var \Prophecy\Prophecy\ObjectProphecy|ClientInterface */
```
Jean85, line 18: With the newest versions, this should work:
```suggestion
    /** @var \Prophecy\Prophecy\ObjectProphecy<ClientInterface> */
```

Jean85 (changes requested):

test/EventListener/MessengerListenerTest.php
ste93cry: I tried and it doesn't work, it gives a bunch of errors the first of which is:

`PHPDoc tag @var for property Sentry\SentryBundle\Test\EventListener\MessengerListenerTest::$client contains generic type Prophecy\Prophecy\ObjectProphecy<Sentry\ClientInterface> but class Prophecy\Prophecy\ObjectProphecy is not generic.`

ste93cry: Nevermind, I updated the `jangregor/phpstan-prophecy` package and now it works!

Jean85 (commented):
Changes look fine for me; we obviously need to fix the issues in the CI...

ste93cry:
I updated the CHANGELOG and added support for a few new options. However, the build on PHP 7.2 keeps failing and I have no idea how to fix it once for all because even if we workaround the issue in the `FrameContextifierIntegration` integration it may arise somewhere else and without changing how we handle the integration of the error handler with the framework I don't see any manouver space

CHANGELOG.md
```diff
@@ -1,87 +1,103 @@
...
-The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
-and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
- - CLI commands registration policy changed to lazy load
+
+- Enable back all error listeners from base SDK integration (#322)
+- Added `options.traces_sampler` and `options.traces_sample_rate` configuration options (#385)
+- [BC BREAK] Removed the `options.project_root` configuration option. Instead of setting it, use a combination of `options.in_app_include` and `options.in_app_exclude` (#385)
+- [BC BREAK] Removed the `options.excluded_exceptions` configuration option. Instead of setting it, configure the `IgnoreErrorsIntegration` integration (#385)
```
Jean85, line 8: Why should we remove this? The bundle is already easily wiring the excluded exceptions into the integration...

composer.json
```diff
@@ -15,13 +24,10 @@
             "email": "alessandro.lai85@gmail.com"
         }
     ],
-    "config": {
-        "sort-packages": true
-    },
```
Jean85, line 20: Why did you remove this?

Jean85 (changes requested):

composer.json
ste93cry: I didn't, I just reordered the content of the file to align the sorting of the sections with the one of the main SDK. It wasn't really important, but I did it anyway as a personal pleasure. Do you want me to revert it back?

CHANGELOG.md
ste93cry: For a bunch of reasons, not all of them are applicable yet:

- The `IgnoreErrorsIntegration` integration may one day accept more than one config option (e.g. the original one in the JS SDK already does it) and I think that it would not be clear to treat such options as if they were options of the SDK
- I would like in the near future to discuss about transforming `sentry.options` to an untyped array node. The reason for this is that we would not need to remap every option of the base SDK in this bundle which is error prone, time consuming and it's a mess once new versions of the SDK adds or deprecate options. In turn, having `excluded_exceptions` as a child of this node would not be correct then because in effect it's not an existing option anymore

Jean85: I'm ok with moving `excluded_exceptions` outside the `option` node, so that we're just giving the user a shortcut for easily configuring the integration.

As for the untyped array, I'm ok too, but we need to force the validation of the values ASAP, like on cache warmup/bundle load.

composer.json
Jean85: Yeah that will maintain the order on the long run.

CHANGELOG.md
ste93cry: >I'm ok with moving `excluded_exceptions` outside the option node, so that we're just giving the user a shortcut for easily configuring the integration.

I have some doubts about this, people may wonder why some integration can be configured like this while others cannot. Wouldn't be better to improve the documentation to explain how to configure the `IgnoreErrorsIntegration` integration?

Jean85: I have two reasons to maintain that:
 * that's the most common-used integration at all
 * the integration is not so straightforward to declare in Symfony, especially in YAML

ste93cry: >the integration is not so straightforward to declare in Symfony, especially in YAML

It would be better to improve the DX of this instead of scattering around the config of the bundle the options of a particular integration :roll_eyes:

Jean85: I don't know if that's feasible without changing a lot how the SDK works..

ste93cry: Shouldn't be enough to declare the integration as a service and then reference it in the `integrations` option?

Jean85: Mh, probably true.

We should start an `UPGRADING` document to reference about those issues.

ste93cry: Done!

UPGRADE-4.0.md
```diff
@@ -0,0 +1,25 @@
...
+
+  ```yaml
+  sentry:
+      integrations:
+          - '@Sentry\Integration\IgnoreErrorsIntegration'
+  
+  services:
+      Sentry\Integration\IgnoreErrorsIntegration:
+          arguments:
+              - [RuntimeException]
```
Jean85, line 24: The array seems wrong here...
Also, having a multiline array here would help a lot:
```suggestion
             $options: 
               ignore_exceptions:
               - \RuntimeException
```

I'm not sure about the syntax, we should try it out.

Jean85 (changes requested):

UPGRADE-4.0.md
ste93cry: >The array seems wrong here...

Right! I fixed it

Jean85 (approved):

PR #387, opened 2020-11-17
Use the hub injected in the constructor of the listeners rather than retrieving it from the SDK singleton

ste93cry:
This PR refactorizes the event listeners to make them use the [`Hub`](https://develop.sentry.dev/sdk/unified-api/#hub) injected into the constructor rather than getting it using the global functions. This eases the unit testing of those classes. Note that by default the injected hub is under the hood an instance of [`HubAdapter`](https://github.com/getsentry/sentry-php/blob/745b3faafdc865ea8de75e0dc61ddfdd36a31d9c/src/State/HubAdapter.php) and thus it works exactly as before by proxing each call to the current hub that is retrieved on-the-fly


ste93cry:
@Jean85 do we really need to make the priority of the listeners configurable? I know it's a really small feature and doesn't have a big impact in the maintenance of the code, but I don't think it's useful since people can simply adjust the priority of their listeners

Jean85:
Listeners need to run after certain stuff, making them configurable makes the installation a lot easier. Requiring users to change their code to install ours is not fine IMHO.

What issues do you have with those?

ste93cry:
No issue at all, I was simply questioning because I never heard anyone needing this and also trying to look around GitHub using the search feature to see who does the same I didn't find much results. I'm totally fine with leaving this as-is if you think it's worth keeping the feature though, I have some doubts about the usefulness but it's just my opinion

Jean85:
The issue was requested in #49 the first time, so there's definitely a need.

ste93cry:
Looking at that PR I think that the main issue was that the listeners at that time were all using the default priority, so people could not inject their own listeners between the Symfony ones (if there was any) and the Sentry ones. You decided to solve the issue by making the priority configurable, which is a way to solve the issue. The other way (it would had required a new major version of course to avoid breaking BC) was to explicitly set the priority. Both ways are fine, as I said I don't really have any issue with keeping things as they are now 😃

phpstan-baseline.neon
```diff
@@ -31,19 +31,19 @@ parameters:
 			path: src/DependencyInjection/SentryExtension.php
 
 		-
-			message: "#^Parameter \\$event of method Sentry\\\\SentryBundle\\\\EventListener\\\\ErrorListener\\:\\:onKernelException\\(\\) has invalid typehint type Symfony\\\\Component\\\\HttpKernel\\\\Event\\\\GetResponseForExceptionEvent\\.$#"
+			message: "#^If condition is always true\\.$#"
```
Jean85, line 34: Al those ignored error should be trimmed down... Maybe in a separate PR.

phpstan.neon
```diff
@@ -1,9 +1,12 @@
...
         - src/
         - test/
-    ignoreErrors:
-         - '/Symfony\\Component\\HttpKernel\\Event\\(GetResponse|FilterController)Event not found.$/'
-includes:
-    - phpstan-baseline.neon
+    excludes_analyse:
+        - test/End2End/App
+    dynamicConstantNames:
+        - Symfony\Component\HttpKernel\Kernel::VERSION
```
Jean85, line 12: This is a very nice fix!

src/EventListener/ConsoleCommandListener.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Handles an error that happened while running a console command.
+     *
+     * @param ConsoleErrorEvent $event The event
+     */
+    public function handleConsoleErrorEvent(ConsoleErrorEvent $event): void
+    {
+        $this->hub->withScope(function (Scope $scope) use ($event): void {
```
Jean85, line 37: This is too reductive: not having the scope configured at command start means that any event manually captured by the user will lack the command context information.

src/EventListener/RequestListener.php
```diff
@@ -13,129 +13,121 @@
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
 
-if (Kernel::MAJOR_VERSION >= 5) {
+if (version_compare(Kernel::VERSION, '4.3.0', '>=')) {
```
Jean85, line 16: Are we sure about this change?

src/EventListener/SubRequestListener.php
```diff
@@ -2,13 +2,13 @@
...
 
-use Sentry\SentrySdk;
+use Sentry\State\HubInterface;
 use Symfony\Component\HttpKernel\Event\FinishRequestEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
 use Symfony\Component\HttpKernel\Kernel;
 
-if (Kernel::MAJOR_VERSION >= 5) {
+if (version_compare(Kernel::VERSION, '4.3.0', '>=')) {
```
Jean85, line 11: Same here..

Jean85 (changes requested):

phpstan.neon
ste93cry: Yep, found out with this PR that PHPStan supports this feature 😄

src/EventListener/RequestListener.php
ste93cry: Yes, take a look at the Symfony code:

[`GetResponseEvent`](https://github.com/symfony/symfony/blob/43688482bc01c40ec421ff9552e1455df460d3f3/src/Symfony/Component/HttpKernel/Event/GetResponseEvent.php#L17) `@deprecated since Symfony 4.3, use RequestEvent instead`
[`FilterControllerEvent`](https://github.com/symfony/symfony/blob/43688482bc01c40ec421ff9552e1455df460d3f3/src/Symfony/Component/HttpKernel/Event/FilterControllerEvent.php#L18) `@deprecated since Symfony 4.3, use ControllerEvent instead`

phpstan-baseline.neon
ste93cry: I have plans to refactorize the remaining tests so I expect a lof of these errors to disappear. However, there are some entries that are here because of PHPStan that doesn't know the code must run with different versions of the vendors and thus some things may be defined in a version and not in another and I don't think we can do anything about this

src/EventListener/ConsoleCommandListener.php
ste93cry: Yes, I'm aware of this. I stopped setting the name of the command as tag outside the error handler because I remembered that I saw months ago some cases where I didn't expect the command name to be logged, however I cannot think as of now that specific case and I have some doubts about me being wrong at that time. I will revert the change for now

src/DependencyInjection/Configuration.php
```diff
@@ -137,14 +137,16 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+            ->defaultValue(128);
+        $listenerPriorities->scalarNode('sub_request')
             ->defaultValue(1);
         $listenerPriorities->scalarNode('request_error')
             ->defaultValue(128);
         $listenerPriorities->scalarNode('console_error')
-            ->defaultValue(128);
+            ->defaultValue(-64);
+        $listenerPriorities->scalarNode('console_terminate')
+            ->defaultValue(-64);
```
Jean85, line 149: Why those priority changes?

ste93cry: Because of these reasons:

- The `console.terminate` listener runs at priority `-64` which is half way between the default value of 0 and the value of the [`Symfony\Component\Console\EventListener\ErrorListener`](https://github.com/symfony/symfony/blob/49eafee03323253328d76ec9cb4c515f94d2f202/src/Symfony/Component/Console/EventListener/ErrorListener.php#L76) listener
- The `console.error` listener runs at priority -64 which is half way between the default value of 0 and the value of the [`Symfony\Component\Console\EventListener\ErrorListener`](https://github.com/symfony/symfony/blob/49eafee03323253328d76ec9cb4c515f94d2f202/src/Symfony/Component/Console/EventListener/ErrorListener.php#L75) listener. Since our listener had a greater priority before, it ran before the Symfony listener and if someone wanted to log somehow the debug messages (e.g. Symfony logs the command that ran) then he couldn't without changing himself the priority of the listener. I chose the new value by simply dividing the priority of the Symfony listener by 2 so that we are in the middle and there is enough "space" for users to push their own listeners both before and after and because I think that the new value is a better default for the reasons explained before
- The `console.command` listener runs at priority `128` for similar reasons than before, I think it's a safer and better default value due to it running now after the [`Symfony\Bridge\Monolog\Handler\ConsoleHandler`](https://github.com/symfony/symfony/blob/49eafee03323253328d76ec9cb4c515f94d2f202/src/Symfony/Bridge/Monolog/Handler/ConsoleHandler.php#L139) listener. Again, I chose this value arbitrarily so that it's in the middle between the default value of `0` assigned by the `EventDispatcher` component and the Symfony listener priority

Jean85 (approved):

Issue #364, opened 2020-09-11
PHP 8 Support


shyim:
Problem 1
    - sentry/sentry-symfony 3.5.2 requires php ^7.1 -> your PHP version (8.0.0-dev) does not satisfy that requirement.

W0rma:
@Jean85 Can we drop older PHP versions - at least for master? This would it make easier to move to PHPUnit 9 which supports PHP 8. Which PHP versions can be dropped? Only 7.1 or 7.2, too since it will reach EOL very soon?

shyim:
I use it already in prod with PHP 8. Had no trouble so far 😅

Jean85:
We should definitely drop 7.1 in master, since it's not supported by the 3.0 SDK: https://github.com/getsentry/sentry-php/blob/b77ff3783060ce3213011ddae369e550ec985dc8/composer.json#L23

cc @ste93cry

stayallive:
I wanted everyone reading to know that this is currently blocked by getsentry/sentry-php#1087, it's being worked on!

p-romain:
Hi, getsentry/sentry-php#1087 is merged, supporting PHP 8 on this package could now be done.

JellyBellyDev:
Go go go php8 compatibility!!! 🚀

Yozhef:
https://github.com/getsentry/sentry-symfony/pull/394

Bilge:
Also waiting on this.

Jean85:
Please note: this is slated for the 4.0 release, which still has a lot of work still pending: https://github.com/getsentry/sentry-symfony/milestone/8

This is due to the fact that getsentry/sentry-php#1087 will be released in 3.x, which will be supported by our future 4.0 only.

Also, we're currently having issues with the build, with unexpected PHPStan failures in #400
If anyone could help us there, we could ship this faster!

Bilge:
Whilst I understand the next major version will support PHP 8, it would greatly ease upgrading if the 3.x branch also supported PHP 8. If it technically does already, adding the tag should be trivial. Having the tag means users can upgrade to PHP 8 and later upgrade sentry-symfony, instead of having to do both at the same time. Therefore having a stable PHP 8 tag now is important, rather than waiting for sentry-symfony 4 release.

ste93cry:
>it would greatly ease upgrading if the 3.x branch also supported PHP 8

[...]

PR #389, opened 2020-11-18
Support PHPUnit 9

W0rma:
One step closer towards PHP 8 support...

Related to https://github.com/getsentry/sentry-symfony/issues/364

* drop support for PHP 7.1
* add support for PHPUnit 9 which itself supports PHP 8

### DIFF
--- a/test/BaseTestCase.php
+++ b/test/BaseTestCase.php
@@ -3,13 +3,26 @@
 namespace Sentry\SentryBundle\Test;
 
 use PHPUnit\Framework\TestCase;
+use Prophecy\PhpUnit\ProphecyTrait;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
 use Symfony\Component\HttpKernel\KernelInterface;
 
-abstract class BaseTestCase extends TestCase
+// Trait is available in phpspec/prophecy-phpunit:2.0 which requires at least PHP 7.3
+if (trait_exists(ProphecyTrait::class)) {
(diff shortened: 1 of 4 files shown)


test/BaseTestCase.php
```diff
@@ -3,40 +3,78 @@
...
-    protected function getSupportedOptionsCount(): int
+// Trait is available in phpspec/prophecy-phpunit:2.0 which requires at least PHP 7.3
+if (trait_exists(ProphecyTrait::class)) {
+    abstract class BaseTestCase extends TestCase
     {
-        return 27;
-    }
+        use ProphecyTrait;
+        
+        protected function getSupportedOptionsCount(): int
```
Jean85, line 19: Are you duplicating the code here? Wouldn't be easier to put an ancestor class with only the trait before this one?

Jean85 (changes requested):
Thanks for the PR! Please take care of the build errors..

test/BaseTestCase.php
W0rma: You are right - fixed. Thank you for the hint

W0rma:
> Thanks for the PR! Please take care of the build errors..

I bumped `symfony/framework-bundle` version to ^3.4.23 which fixes the compatibility issue with older versions of PHPUnit. Tests are green locally. I don't know why the CI build in "prefer lowest" is still failing. Do I need to do something here?

Jean85:
Is still see `Installing symfony/framework-bundle (v3.4.0)` in the build, that's strange...
https://github.com/getsentry/sentry-symfony/pull/389/checks?check_run_id=1422948218#step:7:300

W0rma:
Thank you for the hint - I was able to fix the pipeline

Jean85 (approved):

Jean85:
Thank you for your contribution!

PR #392, opened 2020-11-26
Use symfony kernel

thomas-medelse:


Jean85 (changes requested):
Nice catch, thank you! Can you please fix the build? The code style check is failing.

thomas-medelse:
I don't see the error in my namespace replacement code, can you fix it if you see it?

Jean85:
It's surely due to alphabetical order of the imports. A `make cs` will fix it for you.

thomas-medelse:
The php-cs-fixer does not work on my mac, I tried to correct the error by hand but the test still does not pass.

thomas-medelse:
Thanks

PR #395, opened 2020-12-08
Support PHP8 and update GitHub Actions.

Yozhef:


Yozhef:
@Jean85 could you check?

Jean85:
If the issue is the same as in https://github.com/getsentry/sentry-php/pull/1150, PHPStan has troubles running under PHP 8 due to imprecise function signatures. Try to run it under 7.4 for now, I'll run a separate PR to investigate that (and eventually report bugs upstream).

As for the CI that is failing, you're tripping over 2 additional errors that are caught by Sentry, and both are reported in the CI.

Yozhef:
@Jean85 
I ran PHPStan on 7.4.13 but it, unfortunately, didn't help, but in the local machine all without errors.



What accrued additional problems in the CI - were added Deprecated Message.
Tests (7.2, 3.4.*, Symfony 3.4, disabled) && Tests (8.0, --prefer-lowest, false, false, false)
```
Wrong number of events sent:
b6431231fb48408e84b2d0f110c9eac2: User Notice: This is an intentional notice
###
70a61080c25c4704abc680192ea2a81d: Deprecated: Function libxml_disable_entity_loader () is deprecated
###
4b6ba3285c0b456bb377a11023eaf218: Deprecated: Function libxml_disable_entity_loader () is deprecated
###
```

composer.json
```diff
@@ -19,7 +19,7 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.2",
+        "php": "^7.2|^8.0",
```
localheinz, line 22:
```suggestion
        "php": "^7.2 || ^8.0",
```

Yozhef: Thanks, fixed.

Jean85:
I've rerun CI on the branch and it's still green: https://github.com/getsentry/sentry-symfony/actions/runs/391772879

Could you please split this PR in two? Reorganization of the actions following Sentry-php repo example is good, but it should be done separately so we can find the root cause of this failures.

Yozhef:
@Jean85 yes, thanks.

Jean85:
Closing since #399 got the job done.

PR #397, opened 2020-12-10
Fix: Remove Travis CI badges

localheinz:
This PR

* [x] removes Travis CI badges


Jean85:
Fixed in 3042699 (replacing it with GA badge)

PR #401, opened 2020-12-18
Refactor the bundle configuration and add support for XML format

ste93cry:
With this PR I did a few things, namely:

- Added support for the XML format to the bundle configuration
- Refactored the tests to make them more reliable and improved the code to make it simpler and more readable
- Removed legacy code that was added to support the various features introduced with each `2.x` version
- Refactored how and which container services are registered


Jean85 (commented):
On a first glance, this LGTM, but I need to delve a little more locally, with an IDE.

Unrotunately, the build is still having issues, and I'm making no progress on #400 :cry:

ste93cry:
>Unfortunately, the build is still having issues, and I'm making no progress on #400 :cry: 

Yep I know, some failures of this build are not related to the broken CI anyway. I'm working to fix as much as I can and then I will take a look at the rest. Since the last commit on master is green, something must have changed in the last few versions of PHPStan

src/DependencyInjection/Configuration.php
```diff
@@ -1,211 +1,174 @@
...
+                        ->end()
+                        ->scalarNode('traces_sampler')->end()
+                        ->booleanNode('attach_stacktrace')->end()
+                        ->integerNode('context_lines')->min(0)->end()
+                        ->booleanNode('enable_compression')->end()
+                        ->scalarNode('environment')
+                            ->cannotBeEmpty()
+                            ->defaultValue('%kernel.environment%')
+                        ->end()
+                        ->scalarNode('logger')->end()
```
Jean85, line 72: Shouldn't this be a service?

src/DependencyInjection/SentryExtension.php
```diff
@@ -1,199 +1,157 @@
...
-        }
+        $clientOptionsDefinition = $container->register('sentry.client.options', Options::class);
+        $clientOptionsDefinition->setArgument(0, $options);
 
-        // we ignore fatal errors wrapped by Symfony because they produce double event reporting
-        $ignoreOptions = [
-            'ignore_exceptions' => [FatalError::class],
-        ];
+        $serializerDefinition = $container->register('sentry.client.serializer', Serializer::class);
+        $serializerDefinition->setArgument(0, new Reference('sentry.client.options'));
```
Jean85, line 99: Why this mixed approach in passin the options once as value, twice as reference (created multiple times too)?

src/DependencyInjection/SentryExtension.php
```diff
@@ -1,199 +1,157 @@
...
-            throw new LogicException(
-                sprintf('Missing class "%s", try updating "sentry/sentry" to a newer version.', Handler::class)
-            );
-        }
-
         if (! class_exists(MonologLogger::class)) {
-            throw new LogicException(
-                sprintf('You cannot use "%s" if Monolog is not available.', Handler::class)
-            );
+            throw new LogicException(sprintf('To use the "%s" class you need to require the "monolog/monolog" package.', Handler::class));
```
Jean85, line 150: I'm not sure that's the right package to require in a Symfony app...

src/ErrorTypesParser.php
```diff
@@ -1,81 +0,0 @@
-<?php
-
-namespace Sentry\SentryBundle;
-
-/**
- * Evaluate an error types expression.
- */
-class ErrorTypesParser
```
Jean85, line 8: This is an undocumented BC: before it would work as a string, now you need the `!php/const` format.

But I'm :-1: on this, because in this way I don't think you can use binary operators to obtain something like `E_ALL & ~E_DEPRECATED` as before.

Jean85 (changes requested):

src/DependencyInjection/SentryExtension.php
ste93cry: I think you confused things a bit. What it's being done here is that the `sentry.client.options` service is created and the raw options parsed from the bundle configuration are passed to its constructor. Then, such service is referenced in the constructor of the other two services

src/DependencyInjection/SentryExtension.php
ste93cry: Technically speaking one could also use Monolog without the bundle as it's the only required thing to make our handler work, but for simplicity we may want to suggest the bundle instead of the base library

src/ErrorTypesParser.php
ste93cry: I reverted the change, but I also refactorized the class to make it `static` 👍

README.md
```diff
@@ -159,53 +159,20 @@ composer require pugx/sentry-sdk
...
 
 ## Maintained versions
+
  * 3.x is actively maintained and developed on the master branch, and uses Sentry SDK 2.0;
  * 2.x is supported only for fixes; from this version onwards it requires Symfony 3+ and PHP 7.1+;
  * 1.x is no longer maintained; you can use it for Symfony < 2.8 and PHP 5.6/7.0; 
  * 0.8.x is no longer maintained.
 
-### Upgrading to 3.0
-The 3.0 version of the bundle uses the newest version (2.x) of the underlying Sentry SDK. If you need to migrate from previous versions, please check the `UPGRADE-3.0.md` document.
```
Jean85, line 168: I would leave at least a not with a link to the UPGRADE documents

UPGRADE-4.0.md
```diff
@@ -41,3 +40,123 @@
 
 - Changed the default value of the `sentry.listener_priorities.console_error` configuration option to `-64`.
 - Changed the default value of the `sentry.listener_priorities.console` configuration option to `128`.
+- Changed the default value of the `sentry.listener_priorities.worker_error` configuration option to `50`.
+- Changed the type of the `sentry.options.before_send` configuration option from `scalar` to `string`. The value must always be the name of the container service to call without the `@` prefix.
```
Jean85, line 44: Why did you drop the callable format? Was it harder to maintain?

composer.json
```diff
@@ -21,13 +21,14 @@
...
-        "symfony/http-kernel": "^3.4||^4.0||^5.0",
-        "symfony/security-core": "^3.4||^4.0||^5.0"
+        "php-http/discovery": "^1.11",
+        "sentry/sdk": "^3.1",
+        "symfony/config": "^3.4.43||^4.4.11||^5.0.11",
+        "symfony/console": "^3.4.43||^4.4.11||^5.0.11",
+        "symfony/dependency-injection": "^3.4.43||^4.4.11||^5.0.11",
+        "symfony/event-dispatcher": "^3.4.43||^4.4.11||^5.0.11",
+        "symfony/http-kernel": "^3.4.43||^4.4.11||^5.0.11",
+        "symfony/security-core": "^3.4.43||^4.4.11||^5.0.11"
```
Jean85, line 31: Does Symfony 5.1.0 support PHP 8 correctly?

src/DependencyInjection/Configuration.php
```diff
@@ -1,211 +1,180 @@
...
+                    ->fixXmlConfig('integration')
+                    ->fixXmlConfig('tag')
+                    ->fixXmlConfig('class_serializer')
+                    ->fixXmlConfig('prefix', 'prefixes')
+                    ->children()
+                        ->arrayNode('integrations')
+                            ->scalarPrototype()->end()
+                        ->end()
+                        ->booleanNode('default_integrations')->end()
+                        ->integerNode('send_attempts')->min(0)->end()
```
Jean85, line 51: Does `0` make sense here?

src/DependencyInjection/Configuration.php
```diff
@@ -1,211 +1,180 @@
...
+                            ->scalarPrototype()->end()
+                        ->end()
+                        ->scalarNode('error_types')
+                            ->beforeNormalization()
+                                ->always(\Closure::fromCallable([ErrorTypesParser::class, 'parse']))
+                            ->end()
+                        ->end()
+                        ->integerNode('max_breadcrumbs')
+                            ->min(0)
+                            ->max(Options::DEFAULT_MAX_BREADCRUMBS)
```
Jean85, line 92: Why? Can't I send more?

src/DependencyInjection/Configuration.php
```diff
@@ -1,211 +1,180 @@
...
+                        ->end()
+                        ->variableNode('before_breadcrumb')->end()
+                        ->arrayNode('in_app_exclude')
+                            ->scalarPrototype()->end()
+                            ->beforeNormalization()->castToArray()->end()
+                            ->defaultValue(['%kernel.cache_dir%', '%kernel.build_dir%', '%kernel.project_dir%/vendor'])
```
Jean85, line 98: Should we split this in multiple lines for easier git diff?

src/DependencyInjection/SentryExtension.php
```diff
@@ -1,199 +1,198 @@
...
+            ->setArgument(0, new Reference('sentry.client.options'));
+
+        $representationSerializerDefinition = (new Definition(RepresentationSerializer::class))
+            ->setPublic(false)
+            ->setArgument(0, new Reference('sentry.client.options'));
+
+        $clientBuilderDefinition = (new Definition(ClientBuilder::class))
+            ->setArgument(0, new Reference('sentry.client.options'))
+            ->addMethodCall('setSdkIdentifier', [SentryBundle::SDK_IDENTIFIER])
+            ->addMethodCall('setSdkVersion', [PrettyVersions::getRootPackageVersion()->getPrettyVersion()])
```
Jean85, line 120: This is wrong: this package is NOT the root package when in use.

src/DependencyInjection/SentryExtension.php
```diff
@@ -1,199 +1,198 @@
...
+            // we can save some performance by skipping the rest of the integrations
+            // if the error must be ignored
+            array_unshift($integrations, new Definition(IgnoreErrorsIntegration::class, [['ignore_exceptions' => [FatalError::class]]]));
         }
 
-        $container
-            ->getDefinition(Handler::class)
-            ->replaceArgument('$level', MonologLogger::toMonologLevel($errorHandler['level']))
-            ->replaceArgument('$bubble', $errorHandler['bubble']);
+        return $integrations;
```
Jean85, line 196: Wouldn't a custom sort function be more clear here?

src/ErrorTypesParser.php
```diff
@@ -1,81 +1,69 @@
...
      */
-    private function convertErrorConstants(string $expression): string
+    private static function convertErrorConstants(string $value): string
     {
-        $output = preg_replace_callback('/(E_[a-zA-Z_]+)/', function ($errorConstant) {
-            if (defined($errorConstant[1])) {
-                return constant($errorConstant[1]);
+        $output = preg_replace_callback('/(E_[A-Z_]+)/', static function (array $matches) {
+            if (defined($matches[1])) {
+                return constant($matches[1]);
```
Jean85, line 57: This generates a warning in case of failure. That's bad.

https://www.php.net/manual/en/function.constant.php#refsect1-function.constant-errors

Jean85 (changes requested):

composer.json
ste93cry: Yes, it does since some commit between RC1 and RC2. To be honest I didn't really investigated, I just copied the constraints from the `composer.json` of the main SDK

src/DependencyInjection/Configuration.php
ste93cry: No, as you can see [here](https://github.com/getsentry/sentry-php/blob/490952f8b1c752fa8297af6d7dcdc789b7042320/src/Options.php#L854)

src/DependencyInjection/SentryExtension.php
ste93cry: Right, I didn't think about it when I wrote the code

src/ErrorTypesParser.php
ste93cry: It doesn't because the function call is wrapped around the `defined(...)` check

src/DependencyInjection/Configuration.php
ste93cry: Yes, setting it to that value disables the sending retrial feature. Also, side note, I read somewhere on Discord that there was a discussion regarding the fact that events should simply get lost if the sending is unsuccessful

src/DependencyInjection/Configuration.php
ste93cry: As you prefer, I don't have a strong opinion on this because I don't think it's something that will change in the foreseeable future

src/DependencyInjection/SentryExtension.php
ste93cry: What does sorting have to do with this code 🤔? What I'm doing is to add the `IgnoreErrorsIntegration` integration only if it wasn't already defined in the `integrations` option to avoid overriding the one configured by the user and only if the Sentry error handler is enabled to avoid skipping the log of fatal errors in case the user is using Monolog

UPGRADE-4.0.md
ste93cry: It is indeed harder to handle this syntax across all the configuration formats (e.g. XML) than having a simple string referencing the ID of a service, but that was not my main reason: the main reason is that I wanted to avoid having a syntax specifically of PHP in the configuration files and I don't know any bundle that allows this, so I thought that it was better to step back and just support one kind of syntax that is widely known in the Symfony ecosystem

src/DependencyInjection/Configuration.php
Jean85: It already changed recently due to changes in Symfony folder structure. It could happen again.

src/DependencyInjection/SentryExtension.php
Jean85: Oh right, you're only adding it, not sorting it if present.

src/DependencyInjection/Configuration.php
ste93cry: Yes, but as I said I don't expect it to change anytime soon. Anyway, I split each value into its own line

README.md
ste93cry: I reverted the change and changed the phrase to refer to the 4.0 version since this is what the branch is based on

README.md
```diff
@@ -160,11 +160,16 @@ composer require pugx/sentry-sdk
 
 ## Maintained versions
 
+ * 4.x is actively maintained and developed on the master branch, and uses Sentry SDK 3.0;
  * 3.x is actively maintained and developed on the master branch, and uses Sentry SDK 2.0;
  * 2.x is supported only for fixes; from this version onwards it requires Symfony 3+ and PHP 7.1+;
```
ste93cry, line 165: @Jeans85, now that we are going to release 4.x, what are your plans for `2.x`? I would stop maintaining it and follow the same strategy of the main SDK which is to support only the latest two releases, even if we never documented or said anything about this policy in an official way

Jean85: 2.x works with the 1.x SDK... But I never received an issue on that since 3.0 was released, since it has the same PHP/Symfony requirements.

We could do critical fixes only for both 2 and 3... WDYT?

ste93cry: Considering that `1.x` of the SDK is not maintained anymore and we didn't receive any issue since a long time, I find useless and time consuming to commit supporting `2.x` of this bundle, but I will leave the final decision to you as you are the main contributor here

Jean85: Oh I thought 1.x was still somehow maintained... I'm fine with dropping support for it then, and urgent bugfix for 3.x only :+1:

ste93cry: >Oh I thought 1.x was still somehow maintained

The README still says that that version is maintained for security updates, in reality noone cares anymore about that version. Given that it's two years since the release of the SDK `2.x` and that PHP5 is EOL since when I don't even remember, it's worth in my opinion to officially declare the end of the support for that version and to focus on improving/maintaining the newer versions

Jean85 (approved):

PR #402, opened 2020-12-20
Fix build - PHPStan bootstrap needs to be used

ondrejmirtes:


ste93cry:
Thank you for the resolution of the problem, we continued the work in the original PR

PR #404, opened 2020-12-23
Work Flows separated Symfony integrations.

Yozhef:


ste93cry:
I don't understand what you are trying to achieve here. I would expect only one "CI" workflow that tests all PHP versions for all supported Symfony versions for both lowest and highest dependencies 🤔

Jean85 (changes requested):
I agree, I'm :-1: on this one: with this split approach you either obtain a lot more jobs to run, or you do not test enough combinations of language and Symfony versions.

Yozhef:
thanks for the feedback - I will make changes in the existing one

Issue #384, opened 2020-10-30
How to configure level based on environment variable (monolog integration)


FallDi:
https://github.com/getsentry/sentry-symfony#optional-use-monolog-handler-provided-by-sentrysentry-available-since-320

Is it possible to configure level, based on environment variable?
```
sentry:
    register_error_listener: false # Disables the ErrorListener
    monolog:
        error_handler:
            enabled: true
#            level: error
             level: "%env(LOG_LEVEL)%"
```

Currently it fails with `Incompatible use of dynamic environment variables "LOG_LEVEL" found in parameters.`

Such availability will really useful.

Similar ability was implemented in Monolog bundle recently https://github.com/symfony/monolog-bundle/commit/541e6e56a0b114ddf1dbd08c7392337677b48cae

Following config will be perfect solve such issue
```
parameters:
    sentry.level: '%env(LOG_LEVEL)%'

sentry:
    dsn: "%env(SENTRY_DSN)%"
    options:
        environment: "%kernel.environment%"
    register_error_listener: false # Disables the ErrorListener
    monolog:
        error_handler:
            enabled: true
            level: "%sentry.level%"
```

Jean85:
Closing this since we dropped the functionality in #406. Wrapping the whole Monolog config is too risky and fragile on our side, providing a good out-of-the-box handler should be enough for now.

Issue #337, opened 2020-05-04
Change default hook system to use Monolog


Jean85:
Nicolas, a Symfony Core member, [suggested to change approach](https://github.com/symfony/symfony/issues/36472#issuecomment-616210651) on how to attach Sentry to Symfony.

This requires a new major version since it's a big, impactful change, and still requires #322 as a fallback mechanism.

Jean85:
@B-Galati would you like to help here? Due to your work on https://github.com/B-Galati/monolog-sentry-handler, you're probably an expert on this approach...

tgalopin:
Hello!

This is exactly the reason why I never used this bundle is my own applications before, I'm really glad it's now on the table :) !

I'm posting a comment because I implemented this in most my projects and it may be useful to share it. I think Benoit may even have worked with my implementation on En Marche :) .

I created an open-source project with this implementation, which I didn't share much yet as it's not generic and well-tested enough: https://gitlab.com/citepolitique/oss/sentry.

There are 2 main files in my implementation:

* the handler, actually wiring Sentry with Monolog
* the activation strategy, to avoid logging in Sentry 404 and 403 exceptions

Both are used in Monolog configuration:

```yaml
sentry:
    dsn: '%env(SENTRY_DSN)%'

monolog:
    handlers:
        main:
            type: fingers_crossed
            action_level: error
            activation_strategy: 'citepolitique.sentry.activation_strategy'
            handler: nested
        nested:
            type: group
            members: [file, buffer]
        file:
            type: stream
            path: '%kernel.logs_dir%/%kernel.environment%.log'
            level: debug
        buffer:
            type: buffer
            handler: sentry
        sentry:
            type: service
            id: 'citepolitique.sentry.monolog_handler'
```

This example of configuration is great to see why using Monolog is so powerful: in a few lines of configuration, I added an activation strategy to ignore 403/404, created debug log file, created a debug buffer in memory and sent exception to Sentry with the debug context from the buffer if there is a problem. I can update this behavior easily and chain other handlers if I want to easily.

Jean85:
Thanks for the input @tgalopin! Do you think that that approach is feasible in this bundle? It doesn't seems so to me, but I'm unclear on some inner workings of the Monolog bundle, so I'm probably wrong; more input are more than welcome!

tgalopin:
I'm not sure why it wouldn't be feasible, I just did it in a bundle myself :) . Are there specific reasons blocking you to do it here?

[...]

PR #406, opened 2021-01-06
Remove the config section related to Monolog

ste93cry:
I cannot think of any reason for keeping the `sentry.monolog` configuration option as it doesn't offer any benefit over registering the service by hands. Since changing the Monolog configuration is too hard and cannot be done in an automatic way, the user has anyway to edit the configuration files to add the service to the Monolog handlers stack and at that point having to write the service definition imo is the least. Last, registering the service by hands will make it possible to set environment variables as values of the constructor of the handler, rather than us having to provide workarounds to support this feature in the configuration (#384)


Jean85 (approved):
I agree with this. It will fail hard on update so it's easy to change, and we can later reintroduce it better if we will work on #337

PR #407, opened 2021-01-10
Remove the listener priorities configuration options

ste93cry:
I know we discussed about removing these configuration options in the past, but I strongly believe that this kind of things should not be part of the bundle configuration mainly for these reasons:

- There is already a way to change this kind of things, which is writing a compiler pass or overriding the service
- I think that it's not a common thing to change this, in a lot of years of work with Symfony I rarely if never had to change the priority of a third-party listener
- Exposing this configuration to users puts it on an equal footing with changing any bundle or SDK option, which is false because it affects the inner workings of the bundle itself
- The listeners (for example the tracing listener I would like to implement in the near future) must run in a specific priority which correlates them. Letting users change such priorities will likely break things because they must know that they must also update all related priorities: at that point, you're falling into an advanced use-case which to me means that you also have the knowledge to implement the proper solution


CHANGELOG.md
```diff
@@ -3,18 +3,19 @@
...
-- [BC BREAK] Refactorized the configuration tree and the definitions of some container services (#401)
+- [BC BREAK] Refactorize the configuration tree and the definitions of some container services (#401)
 - Support the XML format for the bundle configuration (#401)
 - PHP 8 support (#399, thanks to @Yozhef)
 - Retrieve the request from the `RequestStack` when using the `RequestIntegration` integration (#361)
 - Reorganize the folder structure and change CS standard (#405)
-- [BC BREAK] Removed the `monolog` configuration option. Instead, register the service manually (#406)
-  
+- [BC BREAK] Remove the `monolog` configuration option. Instead, register the service manually (#406)
+- [BC BREAK] Remove the `listener_priorities` configuration option. Instead, use a compiler pass to change the priority of the listeners (#)
```
Jean85, line 17: Issue number missing.
```suggestion
- [BC BREAK] Remove the `listener_priorities` configuration option. Instead, use a compiler pass to change the priority of the listeners (#407)
```

UPGRADE-4.0.md
```diff
@@ -11,6 +11,7 @@
 - Renamed the `SubRequestListener::onKernelRequest` method to `SubRequestListener::handleKernelRequestEvent`.
 - Renamed the `SubRequestListener::onKernelFinishRequest` method to `SubRequestListener::handleKernelFinishRequestEvent`.
 - Removed the `Sentry\FlushableClientInterface` service alias.
+- Removed the `sentry.listener_priorities` configuration option.
```
Jean85, line 14: We could add an example (before/after) on how to do that, since it's a couple lines of code

composer.json
```diff
@@ -28,6 +28,7 @@
         "symfony/dependency-injection": "^3.4.43||^4.4.11||^5.0.11",
         "symfony/event-dispatcher": "^3.4.43||^4.4.11||^5.0.11",
         "symfony/http-kernel": "^3.4.43||^4.4.11||^5.0.11",
+        "symfony/polyfill-php80": "^1.22",
```
Jean85, line 31: Why are we requiring this directly?

Jean85 (changes requested):

composer.json
ste93cry: PHPStan was complaining on some tests that an anonymous class was passed to a function accepting only a `Stringable` object, so I had to implement that interface interface and thus we need the polyfill

Jean85: Then you should require it as `require-dev` at the very least.

ste93cry: Done 👍

Jean85 (approved):

UPGRADE-4.0.md
futuster: > Letting users change such priorities will likely break things because they must know that they must also update all related priorities

Yeah, an example would be great

PR #410, opened 2021-01-13
feat(Symfony Version) add Symfony Version.

Yozhef:


Jean85 (commented):
I don't see the need for expliciting all those versions. Testing once on every PHP version and at least once for every LTS + latest version seems enough to me.

ste93cry:
If we speak about theory, the variables that we must consider for the matrix are:

- version of Symfony
- version of the dependencies (lowest and highest)
- version of PHP
- OS

Exploding such matrix into builds will generate a lot of them, that's a fact, but it's also the correct way to ensure that everything works well in all scenarios. Any other solution involving less builds is likely better for resources and time, but it will be also more error-prone. I've seen other repos where they have a lot of checks, and I don't think it would be a problem for us to do the same in terms of numbers, and our CI is anyway fast enough to not take ages to complete

Jean85:
Due to the Symfony BC promise, it should be superfluous to test all those versions though.
Also, I'm not sure about OS, since we shouldn't have any OS-dependent handling in this bundle, whereas in the base SDK we have something more low-level that could, and in fact we test it.

Yozhef:
@Jean85 
I think it's important to check now (Symfony 4.4.* / Symfony 5.0. *) together with PHP versions 7.4 and 8.0. in the future the most popular number of projects.

Jean85:
@Yozhef still, the BC promise protects us in that regard: 5.0 is the same as 4.4 without deprecation, and any 5.x will still be compatible with 5.0.

So I would test against:
 * 3.4
 * 4.4
 * 5.x
 * prefer-lowest (which is forcibly 3.4)

This would mean 4 deps versions + 4 PHP versions already (7.2 -> 8.0) or 8 if you count OS variation, for a total of 12 + static analysis... If you add any version in between we're going over 20-30!

Yozhef:
@Jean85 I think that's enough, but ready to listen to what you suggest to change? Now we have:
Symfony 5.0.* PHP 8.0
Symfony 5.0.* PHP 7.4
Symfony 4.4.* PHP 8.0
Symfony 3.4 PHP 7.2
Symfony 4.4.* PHP 7.4
PHP 8.0
PHP 7.4
PHP 7.3
PHP 7.2
PHP 7.2 - lower

Jean85:
I repeat, testing 5.0 on its own in addition to the latest 5.x doesn't make sense, for the same reason for which we don't test for 4.0.

I would test like this:
 * PHP 8.0 (implicitly selects latest SF 5.x)
 * PHP 7.4 (implicitly selects latest SF 5.x)
 * PHP 7.3 (implicitly selects latest SF 5.x)
 * PHP 7.2 (implicitly selects latest SF 5.x)
 * PHP 7.4 SF 4.4
 * PHP 7.4 SF 3.4
 * PHP 7.2 prefer lowest

For a grand total of 7, to double it with OS variation if we want. 14 jobs seems enough to me.

Jean85:
I've re-targeted this to develop to reduce the noise in the diff. It could change again if we merge develop into master due to the preparation of the 4.1 release.

.github/workflows/tests.yaml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: Continuous Integration
```
Jean85, line 1: Changing the name here duplicates the workflow in the web interface; also, a longer name makes it harder to read in the checks recap at the end of a PR page.

.github/workflows/tests.yaml
```diff
@@ -9,66 +9,75 @@ on:
...
-              SYMFONY_DEPRECATIONS_HELPER: disabled
-
-    name: PHP ${{ matrix.php }} tests (${{ matrix.description }})
+          - '7.3'
+          - '7.4'
+          - '8.0'
+        symfony-version:
+          - 3.4.*
+          - 4.4.*
+          - 5.2.*
```
Jean85, line 28: I would prefer it like this:
```suggestion
          - 5.*
```
So we don't have to bump it every six months due to minor releases.

.github/workflows/tests.yaml
```diff
@@ -9,66 +9,75 @@ on:
...
+      - name: Cache Composer dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.directory }}
+          key: ${{ runner.os }}-${{ matrix.php }}-composer-${{ matrix.dependencies }}-${{ hashFiles('**/composer.lock') }}
+          restore-keys: ${{ runner.os }}-${{ matrix.php }}-${{ matrix.dependencies }}-composer-
+
+      - name: Remove Symfony Messenger
+        run: composer remove --dev symfony/messenger --no-update
+        if: matrix.symfony-version == '3.4.*' || matrix.composer_option == '--prefer-lowest'
```
Jean85, line 61: I know that it was like that even before, but why should we remove the messenger on prefer lowest?

.github/workflows/tests.yaml
```diff
@@ -9,66 +9,75 @@ on:
...
           coverage: xdebug
+
+      - name: Setup Problem Matchers for PHPUnit
+        run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+      - name: Determine Composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=directory::$(composer config cache-dir)"
+
+      - name: Cache Composer dependencies
```
Jean85, line 52: This is really verbose. Can we switch to https://github.com/ramsey/composer-install ?

Jean85 (changes requested):
Apart from the requested changes, this produces a big matrix. We should evaluate if all of those are needed, because I don't see enough value in checking all the Symfony versions in each PHP version.

.github/workflows/tests.yaml
ste93cry: >Changing the name here duplicates the workflow in the web interface

Right, too late though :grin:

>a longer name makes it harder to read in the checks recap at the end of a PR page

The name is shorter than the one of the workflow for the static analysis and linting 😏 I'm ready to revert the change if you really think it's an issue, to be honest I don't think so and I find the new name to be better and more self-explicative

.github/workflows/tests.yaml
ste93cry: I have no idea, I just left this as it was before even though I questioned myself why it is the way it is... If there is no reason I agree on removing the latter part of the condition

.github/workflows/tests.yaml
ste93cry: Yes, of course. The same improvement could be made to the `getsentry/sentry-php` repository to keep the GA workflows aligned as much as possible

.github/workflows/tests.yaml
ste93cry: Right, I didn't think about it! Change applied

.github/workflows/tests.yaml
Jean85: I've checked and the text still fits even with a smaller resolution, so we're good.

Jean85:
3.4 is giving us some issues: https://github.com/symfony/symfony/pull/40679 has not been backported there since it's not a security fix, the only way I see to fix this is we should switch to `pcov` here too.

Also, we need to exclude 8.0 + 3.4 because there's no compatible set of components: https://github.com/getsentry/sentry-symfony/pull/410/checks?check_run_id=2336880519

PR #416, opened 2021-01-19
sentry.options.integrations expects the name of the service without the @ prefix

araab:
The UPGRADE-4.0.md contains also a change for the type of the sentry.options.integrations:

> Changed the type of the sentry.options.integrations configuration option from an array of scalar values to an array of string values. The value must always be the name of the container service to call without the @ prefix.


ste93cry (approved):
Ops, it slipped through the code review. Thank you for correcting it!

PR #417, opened 2021-01-20
Added missing schema option

dragosprotung:


ste93cry:
Thank you for the contribution, can you please add an entry to the CHANGELOG?

dragosprotung:
Done

ste93cry (approved):

Jean85:
@ste93cry how did this pass with the tests? Do we miss something there?

dragosprotung:
The integration tests do not validate the XML configuration against the schema.
The field does appear in the tests (see https://github.com/getsentry/sentry-symfony/blob/master/tests/DependencyInjection/Fixtures/xml/full.xml#L29)

ste93cry:
>Do we miss something there?

Yes we do, the `messenger` config node is missing in `tests/DependencyInjection/Fixtures/xml/full.xml` and this caused the tests to not break by fallbacking to the default value for the option

>The integration tests do not validate the XML configuration against the schema.

False, I explained above why the tests didn't break. Try adding the option without your change and Symfony will complain with `[ERROR 1866] Element '{https://sentry.io/schema/dic/sentry-symfony}messenger', attribute 'capture-soft-fails': The attribute 'capture-soft-fails' is not allowed.`. Would you mind opening a new PR to add the missing attribute and fix the tests?

dragosprotung:
Ah, i confused `capture-silenced-errors` with `capture-soft-fails` in the xml fixtures. Opened #420

PR #420, opened 2021-01-21
Test messenger configuration

dragosprotung:
Follow up #417


ste93cry (changes requested):
Can you please add the option to the `DependencyInjection/Fixtures/xml/full.xml` file rather than to the one you changed and set its value to `false` so that in the tests it gets asserted that it is parsed correctly?

dragosprotung:
Added test config for all other formats as well

ste93cry (approved):
Thanks, LGTM!

PR #423, opened 2021-01-22
Initial implementation of sentry tracing with tests

rjd22:
This is my initial experimental implementation of tracing. It is unfinished because it connects quite directly to the dependencies of the project.

# How does it look now
I've run it on my personal pet project. Enjoy!

Screenshot from 2021-01-22 16-16-12


ste93cry:
Thank you for the willingness of contributing! May I ask you to split each feature (the request tracing listener, the Twig tracing listener, etc) into its own PR? This way, it will be both easier to review the code and we can develop them independently without necessarily having to rush for the release of all of them at the same time

rjd22:
@ste93cry Thank you for your feedback. This is an really good idea. I will split them up later.

phpstan.neon
```diff
@@ -8,5 +8,6 @@ parameters:
         - tests
     excludes_analyse:
         - tests/End2End/App
+        - tests/EventListener/Tracing/RequestListenerTest.php
```
rjd22, line 11: You might frown on this but, the test caused so many PHPStan errors because of missing classes it made my head spin.

rjd22:
@ste93cry I splitted off the dbal and twig tracing support and added tests. Testing was quite complex and took me too much time. I'm not planning to invest more of my time into this except for small changes since I'm not getting paid for this.

stayallive:
I'll let @ste93cry and @Jean85 actually comment on the code but looking as a non-symfony dev this looks pretty nice already!

I wanted to at least have thanked you for taking the time to get this started, this is a huge help and very much appreciated. I'm sure Stefano and Alessandro will be able to get this in with this head-start, thanks!

rjd22:
I also made an PR for dbal tracing:
https://github.com/getsentry/sentry-symfony/pull/426

rjd22:
Closing and reopened to trigger build

ste93cry:
Are you willing to rebase this PR so that we can proceed with reviewing it?

rjd22:
@ste93cry yep I will do so later.

rjd22:
@Jean85 did you mean to merge master into this branch? Since it was based on develop.

Jean85:
> @Jean85 did you mean to merge master into this branch? Since it was based on develop.

Crap, that's what I was doing wrong!
Fixing it, in progress!

Jean85:
Ok I force pushed reverting back to a4b5fb3 which was the last commit before my screw up.
I'll attempt to correctly rebase now.

Jean85:
Nope, the right one was 9978c35, and it correctly didn't have any conflicts :sweat: that was messy!

rjd22:
@Jean85 @ste93cry Thanks for your work. I've moved the classes to be ordered similar to the other tracing classes.  And added the changes to the Changelog. I think this one is good to go. When this is merged I will start testing this on a acceptance environment to test the stability and report back the results :).

Asesjix:
Here is a good integration from Jaeger.
You could certainly learn something from the integration.
https://github.com/code-tool/jaeger-client-symfony-bridge

ste93cry:
>Here is a good integration from Jaeger.
>You could certainly learn something from the integration

Thank you. I'm taking inspiration from the Opentelemetry project because I heard about it in the past, but looking at the package you linked it should be on the same wavelength 😃 

>I've moved the classes to be ordered similar to the other tracing classes.

Thank you for contributing to the project. I had no time to leave a code review before you made more changes, I feel sorry for this, but I moved the classes back to the `EventListener` directory because I think that since they are listeners it's better to have all of them under one place only. Instead, the other are specific pieces of integration for third parties that are well suited to be organized under a different folder structure.

>When this is merged I will start testing this on a acceptance environment to test the stability and report back the results :)

Because as of now I mainly used the tests (both unit and e2e) to simulate requests and console commands to check that at least the basics work, this would be really cool!

phpstan.neon
ste93cry: I changed the code to make this unnecessary, generally speaking the less things we exclude from the analysis the better it is 😃

src/Resources/config/schema/sentry-1.0.xsd
```diff
@@ -83,11 +83,16 @@
 
     <xsd:complexType name="tracing">
         <xsd:choice maxOccurs="unbounded">
+            <xsd:element name="request" type="tracing-request" minOccurs="0" maxOccurs="1" />
```
Jean85, line 86: Does it make sense to enable/disable only the request part? Shouldn't this be enabled always when tracing is enables, since it should provide the basic span?

tests/bootstrap.php
```diff
@@ -0,0 +1,23 @@
...
+require_once __DIR__ . '/../vendor/autoload.php';
+
+// According to the Symfony documentation the proper way to register the mocked
+// functions for a certain class would be to configure the listener in the
+// phpunit.xml file, however in our case it doesn't work because PHPUnit loads
+// the data providers of the tests long before instantiating the listeners. In
+// turn, PHP caches the functions to call to avoid looking up the function table
+// again and again, therefore if for any reason the method that should use a mocked
+// function gets called before the mock itself gets created it will not use the
+// mocked methods.
```
Jean85, line 18: Couldn't we use PHPUnit's `setupBeforeClass` hook? So it would also be localized to the test that needs it.

Jean85 (changes requested):

src/Resources/config/schema/sentry-1.0.xsd
rjd22: I would say the same. When you enable tracing I would expect symfony base tracing to enable without extra configuration.

src/EventListener/TracingRequestListener.php
```diff
@@ -0,0 +1,112 @@
...
+     * Gets the HTTP flavor from the request.
+     *
+     * @param Request $request The HTTP request
+     */
+    protected function getHttpFlavor(Request $request): string
+    {
+        $protocolVersion = $request->getProtocolVersion();
+
+        if (str_starts_with($protocolVersion, 'HTTP/')) {
+            return substr($protocolVersion, \strlen('HTTP/'));
```
rjd22, line 107: `substr` is not namespaced and `\strlen` is. Maybe we should make a phpcs check for this. AFAIK. importing them or atleast appending them with `\` saves the compiler some time on the local lookup.

composer.json
```diff
@@ -28,6 +28,7 @@
         "symfony/dependency-injection": "^3.4.44||^4.4.12||^5.0.11",
         "symfony/event-dispatcher": "^3.4.44||^4.4.12||^5.0.11",
         "symfony/http-kernel": "^3.4.44||^4.4.12||^5.0.11",
+        "symfony/polyfill-php80": "^1.22",
```
rjd22, line 31: To be honest I don't feel comfortable to add a polyfill as dependency. This is something that changes global behavior of the engine also for the user of the library. Is it really necessary when we can use `strpos`?

stayallive: Unfortunately at this point it doesn't really matter anymore.

> https://github.com/getsentry/sentry-php/blob/51af981636bbe97fe9271dab76819d02593c3191/composer.json#L38

tests/bootstrap.php
ste93cry: It doesn't work because the data providers in PHPUnit are called before the `setupBeforeClass()` method, therefore the classes will be loaded by the autoloader long before the moment we actually need to mock the time-related functions

src/EventListener/TracingRequestListener.php
ste93cry: >Maybe we should make a phpcs check for this

Actually, this is the result of applying the [`native_function_invocation`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.18/doc/rules/function_notation/native_function_invocation.rst) PHPCS rule

>importing them or atleast appending them with `\` saves the compiler some time on the local lookup

Right, but according to the PHP-CS-Fixer documentation only certain functions benefit from this change, and this is why the `@internal` config is the default value for the rule

composer.json
ste93cry: As said by @stayallive, this doesn't really matter because the polyfill is already required by the main SDK and probably by a lot of other packages as well

>This is something that changes global behavior of the engine also for the user of the library

I don't know what you mean with "changing the global behavior of the engine", but the polyfill only adds functions that are normally available in newer versions of PHP to an older version. There is nothing magic happening here

src/Resources/config/schema/sentry-1.0.xsd
ste93cry: I agree, these were my initial thoughs. However, what if I want to enable tracing for other things but not for the request? Right now, the fact that the request also starts a transaction is just a coincidence

Jean85: Wouldn't the other pieces of tracing break without and underlying new transaction coming from the request?

ste93cry: No, because we always check if a transaction or a span is started, but of course nothing will be traced if neither of them exists. I'm not sure how to proceed, but since starting a transaction is strictly connected with the start of a request, probably you're right in saying that it makes no sense to let users disable the listeners in this specific case

rjd22: @ste93cry Imo I would indeed say the ability of disabling the tracers would most likely not needed.

There is always the option to turn the check around so having it always enabled but giving users the option to disable it? So enabled by default.

ste93cry: >There is always the option to turn the check around so having it always enabled but giving users the option to disable it? So enabled by default.

Yes, on suggestion of @stayallive this is what I'm going to do in an upcoming PR. I'm leaving the code as-is here because it's consistent with how the rest of the PRs got implemented and because doing this kind of change here would be out-of-scope, considering that I have to change all other listeners as well

rjd22:
Btw @ste93cry I've been checking your improvements on my implementation and wanted to thank you personally for how much better this implementation has become.

Thank you!

ste93cry:
FYI, I reverted some of your changes, for example the one that was creating a span for the response and the controller, because I'm not sure we need to go so deep at this moment with tracing and because I'm not sure the events you used were as precise as we would expect. Since I didn't want to postpone merging this PR for too much time, we can go back to the table and improve these aspects further in the next iterations

Jean85 (approved):
Great work everybody!

PR #426, opened 2021-01-24
Add doctrine dbal tracing support that is optional to enable

rjd22:
This is the tracing support for dbal. This also took me half a day figuring out how to make it optional to use and enable. The problem is that there is no clear way to enable tracing because there actually is no hook.

I cheated the system by replacing the symfony DbalLogger that is often used to log queries to monolog. So enabling this will actually disable that functionality. If someone else has an better idea feel free to help out I've been hitting my head on making this work better for 6 hours straight now and could not find any better way.

To use Dbal tracing you will have to enable the following options:

```
doctrine:
    dbal:
        logging: true

sentry:
    tracing:
        dbal_tracing: true
```

Good luck!


rjd22:
The only neat way we could fix the `Class Doctrine\DBAL\Logging\SQLLogger not found.`. Is to install the suggested dependencies for the test. This way we can also unit test the DbalListener.

ste93cry:
>install the suggested dependencies for the test

This is the way 👍

>I cheated the system by replacing the symfony DbalLogger that is often used to log queries to monolog

As far as I know, the Symfony Profiler works the same way by adding a custom SQL logger, which is the public-facing API nearest to where the actual query is really executed. During the past few days I also explored to hook into more low-level points like the driver or the entity manager itself so that we can get more precise tracing of different parts of Doctrine, but for an initial implementation of this feature what you did is more than enough imho

rjd22:
@ste93cry You can indeed wrap the whole driver with composition for more complex tracing en good example of this I found in the following opentracing bundle:

https://github.com/auxmoney/OpentracingBundle-Doctrine-DBAL

rjd22:
@ste93cry I added tests, I feel that this one is GTG

.github/workflows/tests.yaml
```diff
@@ -58,6 +58,7 @@ jobs:
       - run: composer remove --dev symfony/messenger --no-update
         if: matrix.symfony_constraint == '3.4.*' || matrix.composer_option == '--prefer-lowest'
       - run: composer update --no-progress --ansi ${{ matrix.composer_option }}
+      - run: composer suggests sentry/sentry-symfony --list | xargs -i composer require {}
```
ste93cry, line 61: If our code depends on another library, it should be part of our dependencies. Printing a note about a third-party package that enables additional features is a nice thing, but it doesn't mean anything in the context of dependency management

composer.json
```diff
@@ -51,7 +51,8 @@
         "vimeo/psalm": "^4.3"
     },
     "suggest": {
-        "monolog/monolog": "Allow sending log messages to Sentry by using the included Monolog handler."
+        "monolog/monolog": "Allow sending log messages to Sentry by using the included Monolog handler.",
+        "doctrine/dbal": "Trace DBAL queries with Sentry Transactions by using the included DbalListener"
```
ste93cry, line 55: I suggest to change the description to something like `Allow distributed tracing of SQL queries using Sentry`, which is also consistent with the other message

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
```diff
@@ -0,0 +1,33 @@
...
+class ConnectDbalQueryListenerPass implements CompilerPassInterface
+{
+    /**
+     * @return void
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $config = $container->getExtensionConfig('sentry');
+
+        $registerDbalListener = isset($config[0]['register_dbal_listener'])
```
ste93cry, line 22: It would maybe be better to have the definition of this service in the `services.xml` file and remove it if it's disabled: I like to have the most of the things registered in a single place for easy lookup rather than scattering around the codebase the declaration of the services, if not strictly needed

src/DependencyInjection/Configuration.php
```diff
@@ -37,6 +37,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->end()
                 ->end()
                 ->booleanNode('register_error_listener')->defaultTrue()->end()
+                ->booleanNode('register_dbal_listener')->defaultFalse()->end()
```
ste93cry, line 40: Rather than adding a new root-level config option, I was thinking about making a whole new section called `tracing` or `instrumentation` (not sure what the correct term would be, but I believe it depends on the context) with a `boolean` option for each kind of thing we support tracing for

ste93cry (changes requested):
Thank you for the hard work. I did a first pass of code review, my idea is to get this PR merged before reviewing the others so that we don't unnecessarily have to change the others based on agreements we decide here on the initial implementation of things

src/EventListener/Tracing/DbalListener.php
```diff
@@ -0,0 +1,62 @@
...
+use Doctrine\DBAL\Logging\SQLLogger;
+use Sentry\State\HubInterface;
+use Sentry\Tracing\Span;
+use Sentry\Tracing\SpanContext;
+
+/**
+ * Getting the logger, tied into dbal seems extremely hard. Cheating the system a bit by putting it in between the
+ * debug stack logger.
+ */
+final class DbalListener implements SQLLogger
```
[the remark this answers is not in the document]
rjd22: First thank you for your awesome review. I really appreciate working with you.

I was actually thinking about renaming them all to what they are doing, tracing stuff what do you think about moving the Tracing directory out of the EventListener directory and naming them:

DbalSqlLogTracer
KernelEventTracer
TwigTemplateRenderingTracer

This convey better what they are doing namely tracing stuf and what they are tracing. I will also leave more room for new tracers that might use completely different hoofs for tracing.

src/EventListener/Tracing/DbalListener.php
```diff
@@ -0,0 +1,62 @@
...
+    public function startQuery($sql, ?array $params = null, ?array $types = null)
+    {
+        $transaction = $this->hub->getTransaction();
+
+        if (!$transaction) {
+            return;
+        }
+
+        $spanContext = new SpanContext();
+        $spanContext->setOp('doctrine.query');
```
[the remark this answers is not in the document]
rjd22: I had a reason for using a vendor name. Namely being able to discern between different systems.

Sometimes because of migration you have multiple database systems in your software. And being able to see what system is running the query is helpful.

I'm not married with this idea tho. Might a compromise be adding system => doctrine.dbal as data and naming the Op=>db.query?

src/EventListener/Tracing/DbalListener.php
```diff
@@ -0,0 +1,62 @@
...
+        $spanContext->setDescription($sql);
+
+        $this->querySpan = $transaction->startChild($spanContext);
+    }
+
+    public function stopQuery()
+    {
+        if (!$this->querySpan instanceof Span) {
```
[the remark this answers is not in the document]
rjd22: I actually really like using instanceOf because static analyzers like this syntax because the know everything after the if must be an Span.

Would it be okay to leave it like this?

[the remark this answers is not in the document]
ste93cry: >static analyzers like this syntax because they know everything after the if must be an Span

Since the property has the `@var` annotation on it the static analyzers will already complain if you try to assign something else and, so this additional `instanceof` is already "satisfied". Plus, when we will be able to update the minimum required version of PHP to `7.4`, the language itself will provide this guarantee. At this point, the only check that really makes sense to do is guarding that the property is not null, which I find to be more expressive if it's a comparison with the expected value

src/EventListener/Tracing/DbalListener.php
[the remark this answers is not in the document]
ste93cry: >I had a reason for using a vendor name. Namely being able to discern between different systems

I agree, but my thoughs are: what the trace represents is a query (either SQL or a generic database query, can be a MongoDB query or whatever else) while what produces that query is something that should be explained using values persisted in the tags or in the extra data of the span. As a personal opinion, I think that follow the [OpenTracing specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md) (we don't have to blindly follow them since Sentry is not an implementation of them of course) in regards to those fields would be a plus if in the future Sentry will go towards that direction for the APIs

src/EventListener/Tracing/DbalListener.php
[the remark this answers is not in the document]
ste93cry: >what do you think about moving the Tracing directory out of the EventListener directory

Absolutely agree with this, I forgot to ask it in the code review but it would have been part of the next round of requested changes 😬 

>This convey better what they are doing namely tracing stuf and what they are tracing

I would prefer to name the class either `SqlTracerLogger` or `DbalSqlTracerLogger`, because it's effectively a logger that traces things

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
rjd22: I think this is not possible. The problem is that we're replacing a dependency of doctrine. If we do it like you said we end up first replacing it and then removing it also removing the doctrine dependency itself.

Imo allow-listing stuff is always safer than deny-listing in case of inter bundle dependencies.

src/DependencyInjection/Configuration.php
rjd22: Good idea!

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
```diff
@@ -0,0 +1,33 @@
...
+use Sentry\SentryBundle\EventListener\Tracing\DbalListener;
+use Sentry\State\HubInterface;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+class ConnectDbalQueryListenerPass implements CompilerPassInterface
+{
+    /**
+     * @return void
```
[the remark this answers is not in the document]
rjd22: Had to still add the return type for PHPStan. Adding a typehint to the method would result in a interface mismatch.

rjd22:
I applied the changes you requested.

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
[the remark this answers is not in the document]
ste93cry: >Adding a typehint to the method would result in a interface mismatch

This is allowed by PHP and valid according to the https://3v4l.org/RDGjm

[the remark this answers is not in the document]
rjd22: Fixed, thanks for checking!

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
Jean85: Why are we replacing a Doctrine dependency? We should add on top, not interfere.

rjd22: @Jean85 At this moment the DoctrineBundle doesn't have any other integration point for SQL loggers as the set is fixed. This was the most easy and save integration point that would be okay to replace.

I've made a ticket with the DoctrineBundle that explains the problem better but I'm unsure if I have the time and energy to fix this also:

https://github.com/doctrine/DoctrineBundle/issues/1280

src/Tracing/DbalSqlTracingLogger.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function stopQuery()
+    {
+        if (null === $this->span) {
+            return;
+        }
+
+        $this->span->finish();
```
ste93cry, line 63: To avoid holding a reference to the instance beyond what's necessary (the sending of the span as part of the parent transaction), we could reset to `null` the `$span` property

tests/Tracing/DbalSqlTracingLoggerTest.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->listener = new DbalSqlTracingLogger($this->hub);
+    }
+
+    public function testThatDbalStartQueryIgnoresTracingWhenTransactionIsNotStarted(): void
+    {
+        $this->hub->expects($this->once())
+            ->method('getTransaction')
+            ->willReturn(null);
+
+        $this->listener->startQuery('');
```
ste93cry, line 38: I would pass `SELECT * FROM orders` as value (it's just an example query I took from the [OpenTracing specs document](https://github.com/open-telemetry/opentelemetry-specification/blob/67aebda9c3f6ea1007cd964a35122e9a13b8f4b6/specification/trace/semantic_conventions/database.md#mysql)), I know it means nothing but it somehow makes the code feel "real". Also, the same change should be applied everywhere the `startQuery()` method is called

tests/Tracing/DbalSqlTracingLoggerTest.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public function testThatDbalStopQueryIgnoresTracingWhenChildSpanWasNotStarted(): void
```
ste93cry, line 41: Please rename this method to `testStopQueryDoesNothingIfTransactionIsNotStarted`

tests/Tracing/DbalSqlTracingLoggerTest.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        $this->hub->expects($this->once())
+            ->method('getTransaction')
+            ->willReturn(null);
+
+        $this->listener->startQuery('');
+        $this->listener->stopQuery();
+    }
+
+    public function testThatDbalStartQueryAttachesAChildSpanWhenTransactionStarted(): void
```
ste93cry, line 51: There are multiple things we want to assert:

- When `stopQuery()` is called, get span is correctly closed
- When `stopQuery()` is called, the span should have all the information that have been set through its the `SpanContext`

As a last thing, please rename this method to `testStartQueryAttachesChildSpanWhenTransactionIsStarted`

tests/Tracing/DbalSqlTracingLoggerTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $this->listener->startQuery('');
+
+        $spans = $transaction->getSpanRecorder()->getSpans();
+
+        $this->assertCount(2, $spans);
+        $this->assertNull($spans['1']->getEndTimestamp());
+    }
+
+    public function testThatDbalStopQueryFinishesTheChildSpanWhenChildSpanStarted(): void
```
ste93cry, line 68: Once the previous test asserts all the requested things, this one becomes useless and can be deleted

ste93cry (changes requested):

ste93cry:
Before I forget as usual, can you please also add a CHANGELOG entry with a reference to this PR?

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
ste93cry: We should definitely not override the service, this is not acceptable for end users that may be using already a customized logger. We can decorate the existing service with the `Doctrine\DBAL\Logging\ChainLogger` class (if the service itself is not already an instance of it) and then add the logger. We will have to handle a few cases (absence of the addLogger method in newer versions for example), but it should be feasible until someone will step up in `DoctrineBundle` and improve things right there

src/Tracing/DbalSqlTracingLogger.php
```diff
@@ -0,0 +1,65 @@
...
+
+        if (!$transaction instanceof Transaction) {
+            return;
+        }
+
+        $spanContext = new SpanContext();
+        $spanContext->setOp('db.query');
+        $spanContext->setDescription($sql);
+        $spanContext->setData([
+            'db.system' => 'doctrine',
```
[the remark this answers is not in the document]
Jean85: Maybe DBAL is a better indication?

rjd22:
@ste93cry @Jean85 Okay, I'm going to be very honest here. I'm a freelance developer and doing this in my free time. I've been trough 2 review passes already and am burned out. Contributing is not easy and I've now spend 40 of my hours researching making PR's and applying review comments. I'm not getting paid for this.

I might come back later for this, after I recharged...

ste93cry:
First of all, I wanna thank you for the help you gave us until now

>I'm a freelance developer and doing this in my free time

As you know, Sentry is a company, however most of the contributors of this package like you do this in their free time without getting paid. Myself is not getting paid for this, I do it both to contribute back to the community and because I like the product and I feel that working on it is satisfying. Even though, and I know this, often I leave a lot of comments which someone (or a lot of people) may find uninteresting and/or unuseful, I strongly want to keep the code standard and quality of this package the highest I can and this means usually going through many review and changes. Some of these changes are indeed opinionable personal preference and I will try to make my best in the future to improve on this and instead make them myself to not bother people.

> If you have the time to apply the last touches, please do so :).

Sure, once I will have a bit of free time during these days I will happily do it :+1:

rjd22:
@ste93cry thank you for your understanding answer. Could you and @Jean85 make time to check if all the open comments now are all you want changed?

If so I will make one last commit that applies them.

src/Tracing/DbalSqlTracingLogger.php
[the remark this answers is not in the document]
rjd22: @Jean85 @ste93cry If you're okay with it I would leave it as is. Both the doctrine orm and doctrine migrations use the dbal as layer. So naming it Dbal might be confusing for the developer.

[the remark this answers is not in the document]
ste93cry: As I said, I don't personally see what value brings knowing that a certain package started this query, as a user, in case, I would be more interested in knowing to which database system the query is gonna be run on. Anyway, to avoid focusing too much on this, feel free to do what you think is the best thing. Considering we don't know yet what the users expect from the tracing feature, we can always change or improve things later :wink:

ste93cry:
I cannot promise that these changes will be the last, even though I think that after them we should be ok. As I replied above, if you don't have the time to work on this anymore I will anyway step up and end the work because it's indeed something we want to merge and that in the good and in the bad I strongly believe that is a good starting point

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
rjd22: @ste93cry @Jean85 hoe would you solve this then? Sadly there is on other stable integration point. And I feel that rebuilding the DbalLogger on our own is risky because the implementation can change.

But I don't know of a nice way to rename the id of a dependency so that I can wrap it in the ChainSQLLogger to together with ours. And then place it on the dependency id of the DbalSqlLogger.

ste93cry: I pushed some changes to try making the code work with the `ChainLogger` logger. It kinda works, apart from the failing build on PHP 7.2 due to the fact that I based the code on a too recent version of `doctrine/dbal`. Anyway, I am now convinced more that this approach is quite limiting and may not be the best at all for our use-case as we miss a lot of information in the logger (e.g. the database platform). More, having to maintain here the logic to discern what type of SQL logger is already in use and act accordingly to inject our is error-prone, hard and will make maintenance an hell

rjd22: @ste93cry Yep, I agree. The only other option I see if you want more information on the query it wrapping the whole DbalConnection like it done in: https://github.com/auxmoney/OpentracingBundle-Doctrine-DBAL . But I do feel this does add a lot more complexity.

My question would be: Is all this extra complexity worth the added information in your eyes.

src/DependencyInjection/Compiler/DbalSqlTracingLoggerPass.php
```diff
@@ -0,0 +1,90 @@
...
+
+        return null;
+    }
+
+    /**
+     * @param Reference|Definition $loggerDefinition The service definition
+     *
+     * @return Reference|Definition
+     */
+    private function addTracingLoggerToConnectionConfiguration(ContainerBuilder $container, $loggerDefinition)
```
rjd22, line 70: Just a question because I'm unsure myself. But can't we use decoration to wrap the DbalSqlLogger? Wouldn't this simplify this the wrapping by a lot?

https://symfony.com/doc/current/service_container/service_decoration.html

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
ste93cry: >The only other option I see if you want more information on the query it wrapping the whole DbalConnection

I already have a POC based on this PR that works (right now just for the [`Connection::query()`](https://github.com/doctrine/dbal/blob/715a39150cbf961d7ce19b3f1783f6a76a01afe5/lib/Doctrine/DBAL/Driver/Connection.php#L29) method, but it's just a matter of finishing the work). To wrap the driver connection I chose a different approach from the bundle you found because I didn't like at all that they listen to the `postConnect` event to replace the instance of the driver connection using reflection, but the result is the same

>My question would be: Is all this extra complexity worth the added information in your eyes?

Yes, while there are more "moving parts" in wrapping the driver connection, it's a cleaner approach for a few reasons:

- The extension point that allows to wrap the driver connection looks cleaner and more simpler to use than the one of the SQL logger because it already works on the concept of middlewares and therefore we just have to return a wrapped instance: no more code that checks if the logger is an instance of `LoggerChain`, etc
- Since there may be multiple SQL loggers attached to a single connection, the trace may not be enough precise because its timing would take into account things that happen after starting the span depending on the position in the chain in which our logger has been inserted. While it's true that the same thing could happen by wrapping the driver connection, I find this less likely and we're anyway more close to where the query statement is really launched
- Wrapping the driver connection opens more opportunities of having at least a way for accessing additional information which for sure aren't available in the SQL logger, and will allow us to report more precise traces

src/DependencyInjection/Compiler/DbalSqlTracingLoggerPass.php
ste93cry: We cannot because:

- The `LoggerChain` implementation changed in `2.10` and deprecated the old way to add a logger to the chain, so we have to programmatically discern which version of Doctrine DBAL it's running and act accordingly (I didn't do it here, that's the reason of the failing build)
- The whole point is avoiding to wrap an existing `LoggerChain` into another one, as performances may be impacted (even though I think it would be negligible). Another fact for this messy code is that I wanted to try my best to put our own logger as the last in the chain so that the trace is not affected too much by third parties code

src/DependencyInjection/Compiler/ConnectDbalQueryListenerPass.php
rjd22: @ste93cry Awesome, thanks for your hard work! I look forward to seeing the solution you came up with.

src/DependencyInjection/Compiler/DbalTracingPass.php
```diff
@@ -0,0 +1,87 @@
...
+
+        /** @var string[] $connectionsToTrace */
+        $connectionsToTrace = $container->getParameter('sentry.tracing.dbal.connections');
+
+        foreach ($connectionsToTrace as $connectionName) {
+            if (!\in_array(sprintf(self::CONNECTION_SERVICE_NAME_FORMAT, $connectionName), $connections, true)) {
+                continue;
+            }
+
+            if (version_compare(PrettyVersions::getVersion('doctrine/dbal')->getPrettyVersion(), '3.0.0', '>=')) {
```
Jean85, line 49: The pretty version may not be a simple version here. If you start using a non-tagged version of DBAL this check will fail.

src/aliases.php
```diff
@@ -55,3 +65,22 @@ class_alias(FilterControllerEvent::class, RequestListenerControllerEvent::class)
         class_alias(GetResponseEvent::class, SubRequestListenerRequestEvent::class);
     }
 }
+
+if (class_exists(Connection::class)) {
+    $doctrineVersion = PrettyVersions::getVersion('doctrine/dbal')->getPrettyVersion();
+
+    if (version_compare($doctrineVersion, '3.0.0', '<') && !class_exists(Result::class, false)) {
```
Jean85, line 72: Same here... couldn't we check the presence of some class to obtain the same result?

Jean85 (changes requested):

src/aliases.php
ste93cry: Agree, I will check for the existence of the `Doctrine\DBAL\Driver\ResultStatement` class which got removed starting from version `3.0`. This should give us enough confidence that we're on the right version for the test

src/DependencyInjection/Compiler/DbalTracingPass.php
ste93cry: Right, I didn't think about it, even though I think it's unlikely someone uses a development version of this dependency

ste93cry:
I can't believe I finally got the build green! PR should be ready now for the review, I tried it by using a SQLite in-memory database and it worked quite well, although there are probably many things that can be improved. @rjd22 it would be awersome if you could take a look and try it 😃

composer.json
```diff
@@ -51,7 +53,8 @@
         "vimeo/psalm": "^4.3"
     },
     "suggest": {
-        "monolog/monolog": "Allow sending log messages to Sentry by using the included Monolog handler."
+        "monolog/monolog": "Allow sending log messages to Sentry by using the included Monolog handler.",
+        "doctrine/dbal": "Allow distributed tracing of SQL queries using Sentry."
```
Jean85, line 57: Shouldn't we suggest the Doctrine bundle instead? IIRC all this stuff works automagically only with it active...

src/DependencyInjection/SentryExtension.php
```diff
@@ -149,6 +153,24 @@ private function registerMessengerListenerConfiguration(ContainerBuilder $contai
...
 
+    /**
+     * @param array<string, mixed> $config
+     */
+    private function registerTracingConfiguration(ContainerBuilder $container, array $config): void
+    {
+        $isConfigEnabled = $this->isConfigEnabled($container, $config['dbal']);
+
+        if ($isConfigEnabled && !class_exists(DoctrineBundle::class)) {
+            throw new LogicException('DBAL tracing support cannot be enabled as the DoctrineBundle bundle is not installed.');
```
Jean85, line 164:
```suggestion
            throw new LogicException('DBAL tracing support cannot be enabled automatically as the DoctrineBundle bundle is not installed.');
```

src/SentryBundle.php
```diff
@@ -4,9 +4,18 @@
...
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 final class SentryBundle extends Bundle
 {
     public const SDK_IDENTIFIER = 'sentry.php.symfony';
+
+    public function build(ContainerBuilder $container): void
+    {
+        parent::build($container);
+
```
Jean85, line 18: IIRC the parent method is empty
```suggestion
```

src/Tracing/Doctrine/DBAL/TracingDriver.php
```diff
@@ -0,0 +1,142 @@
...
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\VersionAwarePlatformDriver as VersionAwarePlatformDriverInterface;
+use Sentry\State\HubInterface;
+
+/**
+ * This is a simple implementation of the {@see DriverInterface} interface that
+ * decorates an existing driver to support distributed tracing capabilities.
+ * This implementation IS and MUST be compatible with all versions of Doctrine
+ * DBAL >= 2.11.
+ *
```
Jean85, line 23: I would add a reference to the 2.10 implementation with a `@see` tag here

src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
```diff
@@ -0,0 +1,256 @@
...
+use Doctrine\DBAL\Driver\Connection as DriverConnectionInterface;
+use Doctrine\DBAL\Driver\Result;
+use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\ParameterType;
+use Sentry\State\HubInterface;
+use Sentry\Tracing\SpanContext;
+
+/**
+ * This implementation wraps a driver connection and adds distributed tracing
+ * capabilities to Doctrine DBAL.
```
Jean85, line 16: Which version of DBAL uses this? If not all of them, I would write it here

src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * @internal
+     */
+    public const SPAN_OP_TRANSACTION_COMMIT = 'sql.transaction.commit';
+
+    /**
+     * @internal
+     */
+    public const SPAN_OP_TRANSACTION_ROLLBACK = 'sql.transaciton.rollback';
```
Jean85, line 48:
```suggestion
    public const SPAN_OP_TRANSACTION_ROLLBACK = 'sql.transaction.rollback';
```

src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
```diff
@@ -0,0 +1,256 @@
...
+            }
+        }
+    }
+
+    /**
+     * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md
+     *
+     * @return array<string, string>
+     */
+    private function getSpanTags(): array
```
Jean85, line 224: Shouldn't we memoize the result of this function? As far as I can see, the inputs don't change...

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
```diff
@@ -0,0 +1,134 @@
...
+use Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriverMiddleware;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class DbalTracingPassTest extends DoctrineTestCase
+{
+    public function testProcessWithDoctrineDBALVersionAtLeast30(): void
+    {
+        if (!$this->isDoctrineDBALVersion3Installed()) {
```
Jean85, line 21: Should we test for the complete absence of Doctrine? Maybe with a dedicated CI job?

Jean85 (changes requested):

src/SentryBundle.php
rjd22: @Jean85 You never know if something is added later. Best is to always call to parent in these cases if you don't control the Dependency.

ste93cry: This seems a good reason to leave the call as-is. I checked other bundles and even if it's not an indicator that one thing is either correct or not, they do the same

src/DependencyInjection/SentryExtension.php
ste93cry: Adding the word "automatically" implies that if that package is installed the tracing would get enabled by default, which is not true and must not be true

src/Tracing/Doctrine/DBAL/TracingDriver.php
ste93cry: The comment seems clear enough to me and everyone can just go in the `doctrine/dbal` repository, choose the `2.10` tag and see the implementation of the interface at that time. What value does adding such link brings to the reader?

src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
ste93cry: I added the same note written for the `TracingDriver` class that makes clear that the interface must be compatible with all versions `>= 2.10`

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
ste93cry: Yes, definitely! I added a job to the CI to re-run all the tests after removing the optional packages (Messenger component, Doctrine DBAL), maybe we could use the `@group` annotation to avoid re-running the full test suite, but considering that some tests are mixed with other things, I didn't bother investigating the feasability. Also, it may be worth at some point to extract the support of distributed tracing for third parties to another repo/package because right now there is a lot of code to avoid the tests to break entirely if the dependency is not installed and consequently the maintenance cost is rising (just look at how many attempts it took to make it work for all PHP versions and with/without packages installed)

src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
ste93cry: Yes, it makes sense. I've updated the code accordingly

Jean85 (approved):
Code LGTM, great work both of you!

Only thing that I would add is a reference to the new config option.

ste93cry:
Where would you add it? The only place I see that makes sense is the documentation on the official website

Jean85:
I would put the default values with a comment here: https://github.com/getsentry/sentry-symfony#configuration-of-the-sdk

ste93cry:
Are we sure we want to do that? The direction I would like to take as already done for Monolog is moving all documentation besides the install instruction to https://github.com/getsentry/sentry-docs to avoid having to maintain in one place only the docs, so I would like to avoid adding more things in the README

Jean85:
I'm :+1: on moving that stuff to the docs if we leave a clear reference to it. Obviously it should end up in a different PR.

rjd22:
@Jean85 @ste93cry thank you for all the work! It's great to see this one merged. Maybe it would be best to look over the main one after this one and finish with the twig tracing:

https://github.com/getsentry/sentry-symfony/pull/423

VincentLanglet:
Hi @rjd22 Could some documentation of this option be added in the Readme ?

PR #429, opened 2021-01-26
Rename the ConsoleCommandListener class to ConsoleListener

ste93cry:
Small change that deprecates the `Sentry\SentryBundle\EventListener\ConsoleCommandListener` class (and service) in favor of `Sentry\SentryBundle\EventListener\ConsoleListener`. The reason behind it is that the event listener actually handles more than just an event related to the specific command


Jean85 (approved):

Jean85:
Static analysis is failing though.

PR #430, opened 2021-01-26
Implementation of optional sentry tracing with twig

rjd22:
At last ;), the Twig tracing with tests. This also includes the same suggested dependencies fix as the DbalListener. Hopefully git can figure out the merge conflicts :D.

To use Twig tracing you will have to enable the following options:

```
sentry:
    tracing:
        twig:
            enabled: true
```

Good luck!

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -96,6 +96,16 @@ parameters:
 			path: src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
 
 		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:enter\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
+			count: 1
+			path: src/Tracing/Twig/TwigTracingExtension.php
+
+		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:leave\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
+			count: 1
+			path: src/Tracing/Twig/TwigTracingExtension.php
+
+		-
 			message: "#^Class Symfony\\\\Component\\\\HttpKernel\\\\Event\\\\GetResponseForExceptionEvent not found\\.$#"
 			count: 1
 			path: src/aliases.php
@@ -219,4 +229,3 @@ parameters:
 			message: "#^Trying to mock an undefined method convertException\\(\\) on class Sentry\\\\SentryBundle\\\\Tests\\\\Tracing\\\\Doctrine\\\\DBAL\\\\StubExceptionConverterDriverInterface\\.$#"
 			count: 1
 			path: tests/Tracing/Doctrine/DBAL/TracingDriverTest.php
-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -96,16 +96,6 @@ parameters:
 			path: src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
 
 		-
-			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:enter\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
-			count: 1
-			path: src/Tracing/Twig/TwigTracingExtension.php
-
-		-
-			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:leave\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
-			count: 1
-			path: src/Tracing/Twig/TwigTracingExtension.php
-
-		-
 			message: "#^Class Symfony\\\\Component\\\\HttpKernel\\\\Event\\\\GetResponseForExceptionEvent not found\\.$#"
 			count: 1
 			path: src/aliases.php
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -163,6 +163,9 @@ final class Configuration implements ConfigurationInterface
                                 ->end()
                             ->end()
                         ->end()
+                        ->arrayNode('twig')
+                            ->canBeEnabled()
+                        ->end()
                     ->end()
                 ->end()
             ->end();
--- a/src/Tracing/Twig/TwigTracingExtension.php
+++ b/src/Tracing/Twig/TwigTracingExtension.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\SentryBundle\Tracing\Twig;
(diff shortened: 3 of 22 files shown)


ste93cry:
I never looked at the internals of Twig in depth, but looking at how the profiler is integrated in Symfony they extends the `Twig\Extension\ProfilerExtension` class. Does doing that have some advantages? Besides this, are you up to rebase this PR and at least align the main changes following the conventions used in the one for Doctrine DBAL?

rjd22:
@ste93cry I actually did this first but decoupled it later on to lower the dependency and risk of breaking on changes there. There is on real advantages for us to extend it. I will rebase this one later this week to apply the changes of the DBAL integration.

ste93cry:
>I will rebase this one later this week 

Ok, only then I will review the code because doing it now would be useless as a lot of things are gonna change :smiley: 

>I actually did this first but decoupled it later on to lower the dependency and risk of breaking on changes there

I just checked and the class is not marked as `@internal`, so I expect it to not be changed by introducing BC changes between patch/minor versions. Anyway, we have to think broader: what if the user, for some reason, is already using that extension? I find it unlikely as I believe it's mostly used to fill the data for the profiler of Symfony, but you never know... In that case, we would call `enter` and `leave` on the `Twig\Profiler\Profile` class twice :thinking:

rjd22:
@ste93cry AFAIK this should not be a problem. We register our own `ProfilerNodeVisitor`. The `Twig\Extension\ProfilerExtension` also caches calls on `enter` and `leave`. We're not using this data so using the `ProfilerExtension` uses unused memory.

After checking it again this was my reason for not using it. Registering your own extensions in Twig like the way I build it is how it is commonly done.

rjd22:
After checking:

https://github.com/symfony/twig-bundle/blob/667d462b8b28d91e2a44b47712171d52b1a57579/Resources/config/twig.php#L91-L92

I feel unsure if we need to use the ProfilerExtension or not. Symfony registers their own and adds the `Profiler` dependency. Afaik registering our own extension does not interfere with this.

Jean85:
IMHO we should also check if the `ProfilerExtension` isn't bound to non-prod environments: from what I can remember, in Symfony default configs, all that stuff is shut down with `debug: false`, and we can't ask the user to switch on all that stuff!

rjd22:
@ste93cry @Jean85 I rebased it and modified it to how the Dbal tracing was build and added tests. This one is good to go.

phpstan-baseline.neon
```diff
@@ -95,6 +95,16 @@ parameters:
...
+		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:enter\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
+			count: 1
+			path: src/Tracing/Twig/TwigTracingExtension.php
+
+		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:leave\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
+			count: 1
+			path: src/Tracing/Twig/TwigTracingExtension.php
+
```
rjd22, line 107: The rules of PHPStan are conflicting with Psalm. This is actually quite confusing.

src/DependencyInjection/Configuration.php
```diff
@@ -163,6 +163,9 @@ private function addDistributedTracingSection(ArrayNodeDefinition $rootNode): vo
                                 ->end()
                             ->end()
                         ->end()
+                        ->arrayNode('twig')
+                            ->canBeEnabled()
+                        ->end()
```
rjd22, line 168: I've added the enabled tag to make it consistent with Dbal but also allowing us to add options later when needed without causing a breaking change.

phpstan-baseline.neon
```diff
@@ -95,6 +95,16 @@ parameters:
 			count: 1
 			path: src/Tracing/Doctrine/DBAL/TracingDriverConnection.php
 
+		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:enter\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
+			count: 1
+			path: src/Tracing/Twig/TwigTracingExtension.php
+
+		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\Tracing\\\\Twig\\\\TwigTracingExtension\\:\\:leave\\(\\) has parameter \\$profile with no value type specified in iterable type Twig\\\\Profiler\\\\Profile\\.$#"
```
Jean85, line 104: This should be solved by typing it as `Profile<Profile>`, because it's iterable and you get more nested profiles inside. Psalm was probably complaining because you weren't adding any more type information, and PHPStan is requiring the same, it wants more type info.

src/DependencyInjection/Configuration.php
```diff
@@ -163,6 +163,9 @@ private function addDistributedTracingSection(ArrayNodeDefinition $rootNode): vo
                                 ->end()
                             ->end()
                         ->end()
+                        ->arrayNode('twig')
+                            ->canBeEnabled()
```
Jean85, line 167: Good choice! :+1:

src/Tracing/Twig/TwigTracingExtension.php
```diff
@@ -0,0 +1,64 @@
...
+
+        if (null === $transaction || !$profile->isTemplate()) {
+            return;
+        }
+
+        $spanContext = new SpanContext();
+        $spanContext->setOp('twig.template');
+        $spanContext->setDescription($profile->getName());
+
+        $this->events[$profile] = $transaction->startChild($spanContext);
```
Jean85, line 47: I'm not sure about the meaning of this... How does this work?

phpstan-baseline.neon
rjd22: I think I really tried everything. Also what you suggested without being able to solve it.

src/Tracing/Twig/TwigTracingExtension.php
rjd22: Twig Profiles can run out of order. The SplObjectStorage stores the Profile object so we can match it when leave is called for that profile by fetching it by Profile object. Does that explain it?

Jean85: Oh ok perfect! Very clear explanation, thank you.

phpstan-baseline.neon
Jean85: I solved it with 29c0dd7; you were right, it was a bit problematic; Psalm was complaining because it wasn't expecting a template.

phpstan-baseline.neon
Jean85: Solved with 29c0dd7

Jean85 (approved):
I solved the static analysis issues, LGTM now! :+1:

ste93cry (approved):
It looks good, more features can always be added later if we will see there is need. LGTM!

rjd22:
@ste93cry thanks a lot. I will rebase the the last one today :).

VincentLanglet:
Hi @rjd22 Could some documentation of this option be added in the Readme ?

Jean85:
This adds tracing capabilities for Twig templates. For more info about tracing (which got released in 4.1) see https://docs.sentry.io/product/performance/distributed-tracing/

PR #444, opened 2021-02-15
Remove outdated instructions to integrate the bundle with Monolog

ste93cry:
The instructions in the README that explain how to integrate the bundle with Monolog are outdated: we could fix them without too much effort, but I would rather remove them and make the official documentation the only source of truth for this kind of things like it's done in the core SDK (no instructions at all besides how to install the package in the repo). This will also reduce the burden of maintenance because there is just one place that has to be kept up-to-date with the changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,43 +92,6 @@ The parameter `options` allows to fine-tune exceptions. To discover more options
 [the Unified APIs](https://docs.sentry.io/development/sdk-dev/unified-api/#options) options and
 the [PHP specific](https://docs.sentry.io/platforms/php/#php-specific-options) ones.
 
-#### Optional: use monolog handler provided by `sentry/sentry` (available since 3.2.0)
-*Note: this step is optional*
-
-If you're using `monolog` for logging e.g. in-app errors, you
-can use this handler in order for them to show up in Sentry. 
-
-First, enable & configure the `Sentry\Monolog\Handler`; you'll also need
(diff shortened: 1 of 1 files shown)


Jean85 (approved):

alexander-schranz:
@ste93cry I'm curious where this is documented now? I think currently the correct way would be:

```yaml
sentry:
    register_error_listener: false # Disables the ErrorListener
    dsn: "%env(SENTRY_DSN)%"

# monolog configuration for sentry
# see also: https://github.com/getsentry/sentry-symfony/blob/3.5.3/README.md#optional-use-monolog-handler-provided-by-sentrysentry-available-since-320
monolog:
    handlers:
        sentry:
            type: service
            id: Sentry\Monolog\Handler

services:
    Monolog\Processor\PsrLogMessageProcessor:
        tags: { name: monolog.processor, handler: sentry }

    Sentry\Monolog\Handler:
        arguments:
            $hub: Sentry\State\HubInterface
            $level: error
```

but I'm not 100% sure 🙈

ste93cry:
>I'm curious where this is documented now

It's not, we still have to update the official documentation, I'm sorry!

>I think currently the correct way would be

You can find an example in the `UPGRADE-4.0.md` file, but generally speaking yes, how you did it should be fine 😉

alexander-schranz:
@ste93cry Thank you for the response. I created a PR to the docs hope its the correct place: https://github.com/getsentry/sentry-docs/pull/3137/files

PR #448, opened 2021-02-19
fix before_breadcrumb example in upgrade-4.0 doc

BenoitLeveque:
It was using before_send instead of before_breadcrumb in the example


ste93cry (approved):
Thank you for noticing this! LGTM

PR #454, opened 2021-02-27
Fix empty or null DSN not disabling Sentry

### LINKED ISSUE #453: Explicit `null` dsn no longer disables Sentry
### Environment

sentry-symfony 4.0.1 / sentry-php 3.1.5

### Steps to Reproduce

Set the following configuration:
```
sentry:
  dsn: ~ # Or ''
```

### Expected Result

Sentry client should not push any event to the Sentry instance.

### Actual Result

But it does. Due to this particular change, a falsy dsn value will be unset from the configuration: https://github.com/getsentry/sentry-symfony/commit/c95a2bd49525db1cf4da5d36c45f19d9e8cc3912#diff-ebe00144d6cecb5ca8a2860883e01147f42bbed2b12a3a7f44915908a4829faaR30

As it is unset, it is not forwarded to the Sentry Client options and that causes sentry-php to fallback to the `SENTRY_DSN` defined in the environment, as that is the default: https://github.com/getsentry/sentry-php/blob/master/src/Options.php#L693

Either the documentation here should be updated (and it should be added to the 4.0 upgrade notes) that you need to disable it using some other method (I do not know which one), or the value should not be unset in the configuration.

ste93cry:
Fixes #453. During the refactoring of the `Configuration` class of the bundle, for some reason I thought that when the DSN value is an empty string or a `null` value it should have been removed from the options array, probably because I forgot that the default value in the base SDK is the `SENTRY_SDK` environment variable and not a `null` value. This causes a big problem when you explicitly set the option to one of those values (for example in the test or development environment) and the environment variable is also set because the option is not passed to the core SDK and Sentry remains enabled.


Jean85 (changes requested):
Is there any way to test this?

ste93cry:
Technically speaking yes, we could add a test that assert that given a certain value of the DSN option it is kept, practically I don't see any value in doing it because there is no code to test

Jean85:
That's still a regression test. It's a test of a functionality, the fact that you had to delete code to make it work doesn't make the bug disappear out of thin air.

ste93cry:
The point is that as I already said there is nothing to test: the fact that a value (regardless of what is it) set for the DSN option is passed down to the `Options` object is already handled [here](https://github.com/getsentry/sentry-symfony/blob/b4ce11276f161a263a80f9554bb6a1819acd4563/tests/DependencyInjection/SentryExtensionTest.php#L173), and since there isn't anymore the code that treats the `null` value differently, the existing test method should inherirently also test this case

Jean85:
That's not the full picture. 

Here you have a side effect from the client hijacking the bundle's configuration. I would imagine a test where I set the `SENTRY_DSN` env variable, set `null` in the bundle and expect `null` in the options. That test would fail on master now, but be green in this PR.

ste93cry:
>I set the `SENTRY_DSN` env variable, set `null` in the bundle and expect `null` in the options

In `getsentry/sentry-php` we test that when no value is set for the DSN we fallback to the environment variable and, if it doesn't exists, then we fallback to `null`. In this bundle instead, we test that the value gets passed as-is to the constructor of the `Options` object. From there on, what happens is business of the other package. Does it makes sense?

Jean85 (approved):
Ok I'm fine with it, let's ship it

PR #457, opened 2021-03-03
Allow explicit null value for dsn configuration option

bobvandevijver:
The fix in #454 missed this particular line: when the `dsn` configuration option was set to `null`, it was still not forwarded to the actual Sentry options.

Setting it to an empty string does work in 4.0.2, so that can be used as temporary workaround.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+## 4.0.3 (2021-03-03)
+- Fix regression from #454 for `null` value on DSN not disabling Sentry (#457)
+
 ## 4.0.2 (2021-03-03)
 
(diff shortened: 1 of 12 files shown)


Jean85 (changes requested):
Wow! Thank you for the very quick report and fix!

We should require tests here then, to avoid another regression on this issue... Can you provide them? Otherwise I'm happy to take over.

bobvandevijver:
@Jean85 I tried to make something fitting. I added tests for three supported falsy values (according to the https://github.com/getsentry/sentry-php/blob/master/src/Options.php#L822). Both the `''` and `false` values pass without my fix, but the null check fails. But, they all pass with the small fix in this PR 👍🏻

Jean85 (approved):
I've added a changelog entry, thank you!

PR #460, opened 2021-03-06
Add support for distributed tracing when running a console command

ste93cry:
With this PR the distributed tracing feature is activated when running a console command. If no transaction exists it starts one, otherwise it starts a span that is child of the current one. This last thing can happen if tracing has been started manually before the invocation of our listener or if the command is being executed in the context of a HTTP request where the `TracingRequestListener` already did its job

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add support for distributed tracing of Symfony request events (#423)
 - Add support for distributed tracing of Twig template rendering (#430)
 - Add support for distributed tracing of SQL queries while using Doctrine DBAL (#426)
+- Add support for distributed tracing when running a console command (#455)
 - Added missing `capture-soft-fails` config schema option (#417)
 - Deprecate the `Sentry\SentryBundle\EventListener\ConsoleCommandListener` class in favor of its parent class `Sentry\SentryBundle\EventListener\ConsoleListener` (#429)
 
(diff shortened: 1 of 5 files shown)


Jean85 (approved):
LGTM :+1:  good job on thinking about the HTTP context

stayallive:
Dis you take `console.command` from another SDK or made that up yourself? 

Checking to see if we need to double check the `op` names were using here to make sure we keep it generally in line with expected values.

ste93cry:
No I didn't, I don't think there is any other SDK (maybe Laravel?) that has tracing support for the console commands

stayallive:
I think none have. So let's check with @HazAT what is expected and/or console commands should even be added (outside the HTTP context) at the moment (I'll ask this in Discord too to get some attention).

ste93cry:
>So let's check with @HazAT what is expected and/or console commands should even be added

We definitely want them, otherwise anything running in the context of a console command (SQL queries, etc) will not work :sweat_smile: This work is basically mirroring what has been done for starting a transaction when an HTTP request is handled

>Checking to see if we need to double check the op names were using here to make sure we keep it generally in line with expected values

I think that for some things Sentry could also find a standard, for others probably not. I would not bother so much with this, for example our SDK is (I believe) the first one that is tracing SQL queries at low-level, which has a great value, and the last thing I want is to be locked with some generic name just because some other SDK introduced them before 😉

PR #461, opened 2021-03-06
Fix failing unit tests after Symfony issue #38019 bugfix

ste93cry:
Unit tests for the `TracingRequestListener` class are failing after the release of the fix for symfony/symfony#38019. Since PHP 5.4, the `REQUEST_TIME_FLOAT` element is always present in `$_SERVER`, but Symfony was not setting the value when creating the request object manually, therefore we had to do it ourselves. Now that this is unnecessary, we can cleanup the code


Jean85 (changes requested):
Ugh. seems we have a fatal on 3.4, where the fix is missing

ste93cry:
Yes I know, the bug is being investigated in symfony/symfony#40390

Jean85:
Issue seems to have been fixed with https://github.com/symfony/symfony/pull/40494, should we wait for a tag?

ste93cry:
Yes, I talked privately with Nicolas and he said me that the new version should be released before the end of the month. The alternative in the meantime would be to fix the `composer.json` to restrict the installed version to a previous version, but I don't think it's worth losing time in doing it by this time

Jean85:
Roger. We need to keep in mind that this same problem is breaking the CI in 3.5.x too, where we have a couple of PR to release as a patch.

ste93cry:
Symfony `5.2.6` has been released and build is now green 🎉

Jean85 (approved):

PR #466, opened 2021-03-13
Enable the distributed tracing instrumentation by default

ste93cry:
As asked by @stayallive, this PR enables all the tracing instrumentation by default. Since the `traces_sample_rate` option can be configured using an environment variable, we have no way to know ahead of time if the tracing will be enabled or not: this has the consequence that we have to keep all services registered in the container and active with the negative note that the performance of the application will be slightly affected. I don't know how much loss we we're talking about, but some instrumentation is deeply integrated with low-level things like the driver's connection of the DBAL so I expect at least a bit.

### DIFF
--- a/src/DependencyInjection/Compiler/DbalTracingPass.php
+++ b/src/DependencyInjection/Compiler/DbalTracingPass.php
@@ -31,15 +31,19 @@ final class DbalTracingPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container): void
     {
-        if (!$container->hasParameter('doctrine.connections')) {
+        if (!$container->hasParameter('doctrine.connections') || !$container->getParameter('sentry.tracing.dbal.enabled')) {
             return;
         }
 
-        /** @var string[] $connections */
-        $connections = $container->getParameter('doctrine.connections');
-
         /** @var string[] $connectionsToTrace */
         $connectionsToTrace = $container->getParameter('sentry.tracing.dbal.connections');
+
+        /** @var array<string, string> $connections */
+        $connections = $container->getParameter('doctrine.connections');
+
+        if (empty($connectionsToTrace)) {
+            $connectionsToTrace = array_keys($connections);
+        }
 
         foreach ($connectionsToTrace as $connectionName) {
             if (!\in_array(sprintf(self::CONNECTION_SERVICE_NAME_FORMAT, $connectionName), $connections, true)) {
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -8,6 +8,7 @@ use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use Jean85\PrettyVersions;
 use Sentry\Options;
 use Sentry\SentryBundle\ErrorTypesParser;
+use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
@@ -150,17 +151,16 @@ final class Configuration implements ConfigurationInterface
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->arrayNode('dbal')
-                            ->canBeEnabled()
+                            ->{class_exists(DoctrineBundle::class) ? 'canBeDisabled' : 'canBeEnabled'}()
                             ->fixXmlConfig('connection')
                             ->children()
                                 ->arrayNode('connections')
-                                    ->defaultValue(class_exists(DoctrineBundle::class) ? ['%doctrine.default_connection%'] : [])
                                     ->scalarPrototype()->end()
                                 ->end()
                             ->end()
                         ->end()
                         ->arrayNode('twig')
-                            ->canBeEnabled()
+                            ->{class_exists(TwigBundle::class) ? 'canBeDisabled' : 'canBeEnabled'}()
                         ->end()
                     ->end()
                 ->end()
--- a/src/DependencyInjection/SentryExtension.php
+++ b/src/DependencyInjection/SentryExtension.php
@@ -24,6 +24,7 @@ use Sentry\SentryBundle\Tracing\Twig\TwigTracingExtension;
 use Sentry\Serializer\RepresentationSerializer;
 use Sentry\Serializer\Serializer;
 use Sentry\Transport\TransportFactoryInterface;
+use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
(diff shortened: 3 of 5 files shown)


src/DependencyInjection/Compiler/DbalTracingPass.php
```diff
@@ -31,16 +31,20 @@ final class DbalTracingPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container): void
     {
-        if (!$container->hasParameter('doctrine.connections')) {
+        if (!$container->hasParameter('doctrine.connections') || !$container->getParameter('sentry.tracing.dbal.enabled')) {
```
Jean85, line 34: Shouldn't we check for the whole tracing to be disabled too?

Jean85 (changes requested):
I'm not sure we should turn this on everywhere, it's a new feature and not the main one (which is error tracing) so not all users are expecting that. I don't like so much the performance hits...

Also, weren't we talking about a main switch?

stayallive:
I'm not familiar enough with how Symfony "boots" but in Laravel we register with the container (just registering classes no actual `new`ing classes) and in the boot sequence of Laravel we inject our middleware and other changes to inject tracing into the framework.

> https://github.com/getsentry/sentry-laravel/blob/0c932f2ba38acca9988d2585dc60dd358cd15ab1/src/Sentry/Laravel/Tracing/ServiceProvider.php#L16-L34

For Laravel this is "just in time" and we have access to the config (merged with `.env` and whatnot) to check if we should enable tracing.

In various places we make sure we do as little as possible when there is no current span set:

> https://github.com/getsentry/sentry-laravel/blob/0c932f2ba38acca9988d2585dc60dd358cd15ab1/src/Sentry/Laravel/Tracing/EventHandler.php#L65-L70

The few bits and bobs we do register (although tracing might be disabled or sampled out) are very lean and add little to no overhead to the request.

I'm not sure if this helps or is useless info but this is how we handle it in Laravel to be ready for tracing but do as little when tracing turns out to be disabled.

The reason I asked for a change like this is because if the user enables tracing all "integrations" (DBAL, HTTP requests etc.) should be enabled by default (it is of course still fine to have a config option to disable them) so there is minimal config to tweak to start with tracing.

Jean85:
> The reason I asked for a change like this is because if the user enables tracing all "integrations" (DBAL, HTTP requests etc.) should be enabled by default (it is of course still fine to have a config option to disable them) so there is minimal config to tweak to start with tracing.

I agree with this approach, but I'm not sure if what this PR is accomplishing right now is this. IMHO, the end user should obtain the bundle config with these default values:
```
sentry:
  tracing:
    enabled: false
    dbal:
      enabled: true
    twig:
      enabled: true
```
In this way, simply by doing
```
sentry:
  tracing:
    enabled: true
```
all the instrumentation would be ready to go.
Right now this `sentry.tracing.enabled` options is missing, and IMHO we can't enable all this stuff with that option present and defaulted to `false`.

ste93cry:
>In various places we make sure we do as little as possible when there is no current span set:

We do the same here

>The few bits and bobs we do register (although tracing might be disabled or sampled out) are very lean and add little to no overhead to the request

That's what I was saying: I have honestly no clue (I didn't profile a sample app) about how much performance loss people will have with the tracing features enabled, I expect at least a small bit because of more things running and more abstractions (mainly in the DBAL) but it could very well be unnoticeable, considering that PHP itself is not the best language when talking about performances 😃 

>The reason I asked for a change like this is because if the user enables tracing all "integrations" (DBAL, HTTP requests etc.) should be enabled by default

I think we all agree that this makes totally sense

>Right now this `sentry.tracing.enabled` options is missing, and IMHO we can't enable all this stuff with that option present and defaulted to `false`

It doesn't make any sense to add one more option to control if tracing is enabled or not because by definition it is enabled as soon as `traces_sample_rate` is a value greater than 0. This is why there isn't another global switch 😅 If we want to introduce a flag specifically for that, people will have to handle two options to enable the tracing which I'm pretty sure is not something that we want. But as I tried to explain, since this kind of configuration is likely to be tuned using environment variables, we cannot know at compile time what the value will be, so we have to register all services and enable all of them, and then just do nothing when they run if at runtime we see that tracing is disabled. How it's done right now will also make it possible to enable tracing on the fly, which would not if no service was instantiated. Of course, if people decide that they want to disable a certain instrumentation, then they cannot complain to have it working should they enable tracing later at runtime

TL;DR: it's working like in Laravel: all services are registered in the DI container by default, but they are basically noop if tracing is disabled. Still, they are there so there may be some performance hit in some critical paths like the connection of Doctrine

Jean85:
But there's a profound difference between how Laravel handles this noops and how we're doing it, as @stayallive correctly pointed out:

 > [...] in Laravel we register with the container (just registering classes no actual `new`ing classes) [...]

In Symfony instead we have a compiled DI container, so we're injecting at least 3 listeners and a decoration of the DB connections just out the door, without the user asking for it. That's why I think it is necessary to add a switch on top of the `traces_sample_rate`: I want to be able to remove all those injected services because I don't want them!

Profiling has a non negligible impact on performance because it lies in all the hot paths of the applications, and I don't want this bundle to be blamed for slowing down the user's app. Profiling is an optional feature of Sentry, I don't expect our userbase to use it in more than 10% of their time, so I'm still :-1: and for adding the `sentry.tracing.enabled` main switch.

If you think that having two settings is wrong, we could make the configuration bail if no `traces_sample_rate` (even with `0`) is set when `sentry.tracing.enabled` is set to `true`, so we could guide the user in the best possible way.

ste93cry:
>I want to be able to remove all those injected services because I don't want them!

I absolutely agree with you, but I can already see users setting `sample_traces_rate` to a value greater than `0` and then complaining because it doesn't work just because they didn't set `sentry.tracing.enabled` to `true`. Keeping in sync the two options would be even more complicated, and in reality we have the same exact problem of environment variables with the latter. As I assume that users will probably want to decide at runtime if tracing must be enabled, if we don't register the services it will not work

stayallive:
Can't you just add the tracing option and not have it in the template supplied to the users?

This way you can easily set `tracing.enabled = false` but you don't have to remember to turn it on?

This way you can completely disable tracing but have the default being it enabled. In Laravel we have a seperate "service provider" you can de-register to completely remove all Tracing bindings, if that is not an EZ option in Symfony having a "hidden" (but documented) option is maybe a good compromise?

Jean85:
@stayallive so you mean putting `sentry.tracing.enabled` by default to `true`, so it's useful only in the case of wanting to strip everything? That could be an interesting solution...

stayallive:
Yeah, you can then "disable" (more de-activate) tracing by setting `traces_sample_rate` to `0.0` or return `0.0` from `traces_sampler` and if you really don't want any of it running in your app you can set `sentry.tracing_enabled` (or maybe even `sentry.tracing_activated` or `tracing_loaded`) which defaults to true to completely strip any tracing code from being initialized/registered or whatever.

This sounds like it could be a nice middle ground to allow removing tracing code if you really wanted?

ste93cry:
>This way you can easily set `tracing.enabled = false` but you don't have to remember to turn it on?

If we want to go this way, we have to mention in the docs will be that this bundle is gonna hook the instrumentation things by default and so it's likely that there will be a performance penalty right after the installation: if people are not interested to this feature, it's better to set the sentry.tracing.enabled option to false as first thing

Jean85:
Yes, but I don't mind it: if we already have the solution ready, we can point anyone to that easy fix, and be blameless.

I would preferred not turning it on by default, but I can understand Sentry wanting to make the feature adoption smoother.

ste93cry:
Superseded by #467

PR #467, opened 2021-03-16
Add global switch to allow disabling all instrumentations for the tracing feature

Jean85:
As suggested by @stayallive in https://github.com/getsentry/sentry-symfony/pull/466#issuecomment-800084751


Jean85:
I'm not sure if this needs more tests, otherwise it's ready for review.

tests/DependencyInjection/SentryExtensionTest.php
```diff
@@ -302,7 +305,20 @@ public function dsnOptionIsSetOnClientOptionsDataProvider(): \Generator
...
+    public function testAllIntegrationsAreRemovedWhenTracingIsDisabled(): void
+    {
+        $container = $this->createContainerFromFixture('tracing_disabled');
+
+        $this->assertFalse($container->hasDefinition(TracingRequestListener::class));
+        $this->assertFalse($container->hasDefinition(TracingSubRequestListener::class));
+        $this->assertFalse($container->hasDefinition(TracingConsoleListener::class));
+        $this->assertFalse($container->hasDefinition(TracingDriverMiddleware::class));
+        $this->assertFalse($container->hasDefinition(ConnectionConfigurator::class));
+        $this->assertEmpty($container->getParameter('sentry.tracing.dbal.connections'));
```
ste93cry, line 317: A really small change, just for personal pleasure: can you please move this line as last one 🙏?

ste93cry (changes requested):

stayallive (approved):
If I'm understanding it correctly this seems like a great solution 👌

ste93cry (changes requested):
@Jean85 you forgot to enable all tracing instrumentations, which we want to do if we want to keep `sentry.tracing.enabled = true` by default. I started this work in #466, so feel free to cherry-pick the commits from there or merge this PR into mine and then I will make the final changes to tie everything together

Jean85:
@ste93cry I would prefer not to mix the two PRs to avoid making the review harder. I've switched them on by default, you can go ahead afterward with your PR with just fixing the small conflict.

Jean85:
Ok, a full-red CI has shown me why that wasn't a good idea... Let's merge #466 first then, and then we can come back here.

Jean85:
Ok, scratch that: I've merged #466 + rebased, I'll address the comments coming from that PR and we should be good to go!

Jean85 (commented):
@ste93cry @stayallive I've merged #466 into this one, you can review it as a whole now.

I've left a few comments to ease the review.

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
```diff
@@ -129,6 +152,16 @@ private function createContainerBuilder(): ContainerBuilder
             ->register(ConnectionConfigurator::class, ConnectionConfigurator::class)
             ->setPublic(true);
 
+        $container->setParameter('sentry.tracing.enabled', true);
+        $container->setParameter('sentry.tracing.dbal.enabled', true);
```
ste93cry, line 156: This line is repeated twice

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
```diff
@@ -23,7 +23,6 @@ public function testProcessWithDoctrineDBALVersionAtLeast30(): void
         }
 
         $container = $this->createContainerBuilder();
-        $container->setParameter('doctrine.connections', ['doctrine.dbal.foo_connection', 'doctrine.dbal.bar_connection', 'doctrine.dbal.baz_connection']);
```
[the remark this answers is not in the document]
ste93cry: I'm not sure this is a good idea because this way all tests will share the same connections and the same container parameters, but not all tests need them at all or need them configured in a particular way. My fear is that in the future we may get into the trap that we cannot evolve the test methods independently as easy as of now

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
```diff
@@ -90,30 +89,54 @@ public function testProcessWithDoctrineDBALVersionLowerThan30(): void
 
         $connection1 = (new Definition(Connection::class))->setPublic(true);
         $connection2 = (new Definition(Connection::class))->setPublic(true);
+        $connection3 = (new Definition(Connection::class))->setPublic(true);
```
ste93cry, line 92: Why you added a third connection?

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
```diff
@@ -90,30 +89,54 @@ public function testProcessWithDoctrineDBALVersionLowerThan30(): void
...
+        yield [$container];
 
-        $container->compile();
+        $container = $this->createContainerBuilder();
+        $container->setParameter('doctrine.connections', []);
+
+        yield [$container];
 
-        $this->assertEmpty($container->getDefinition('doctrine.dbal.foo_connection.configuration')->getMethodCalls());
+        $container = $this->createContainerBuilder();
```
ste93cry, line 136: After these changes, this test case is the same as the first one of the data provider. Did you meant to test `sentry.tracing.enabled` in one of them?

ste93cry (changes requested):

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
```diff
@@ -23,7 +23,6 @@ public function testProcessWithDoctrineDBALVersionAtLeast30(): void
         }
 
         $container = $this->createContainerBuilder();
-        $container->setParameter('doctrine.connections', ['doctrine.dbal.foo_connection', 'doctrine.dbal.bar_connection', 'doctrine.dbal.baz_connection']);
         $container->setParameter('sentry.tracing.dbal.connections', ['foo', 'bar', 'baz', 'qux']);
```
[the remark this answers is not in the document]
ste93cry: Yes, correct, if no connection exists with that name we will simply skip it. Maybe we should be more strict in this regards?

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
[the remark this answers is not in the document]
ste93cry: I'm not sure this is a good idea because this way all tests will share the same connections and the same container parameters, but not all tests need them at all or need them configured in a particular way. My fear is that in the future we may get into the trap that we cannot evolve the test methods independently as easy as of now

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
Jean85: Because by default now we have 3 connections in the container, and the test would otherwise fail.

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
Jean85: Yes, exactly. My mistake, I changed the first one.

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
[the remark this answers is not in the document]
Jean85: That would probably be good, since it would blow up on cache clear, not in prod :+1:

tests/DependencyInjection/Compiler/DbalTracingPassTest.php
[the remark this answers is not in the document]
ste93cry: As discussed on Discord, let's leave this as-is and postpone any change to the moment we will need them

ste93cry (approved):

PR #477, opened 2021-04-04
Add distributed tracing support for the Symfony Cache component

ste93cry:
Here we go with another instrumentation, this time for the [Cache](https://symfony.com/doc/current/components/cache.html) component. Its implementation is quite simple, and it resembles the one of the DBAL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Add support for distributed tracing of Twig template rendering (#430)
 - Add support for distributed tracing of SQL queries while using Doctrine DBAL (#426)
 - Add support for distributed tracing when running a console command (#455)
-- Added missing `capture-soft-fails` config schema option (#417)
+- Add support for distributed tracing of cache pools (#)
 - Deprecate the `Sentry\SentryBundle\EventListener\ConsoleCommandListener` class in favor of its parent class `Sentry\SentryBundle\EventListener\ConsoleListener` (#429)
 - Lower the required version of `symfony/psr-http-message-bridge` to allow installing it on a project that uses Symfony `3.4.x` components only (#480)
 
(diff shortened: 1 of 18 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 - Add support for distributed tracing of Twig template rendering (#430)
 - Add support for distributed tracing of SQL queries while using Doctrine DBAL (#426)
 - Add support for distributed tracing when running a console command (#455)
-- Added missing `capture-soft-fails` config schema option (#417)
```
Jean85, line 9: Is this removal intentional or has it been lost during rebase?

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -46,6 +47,9 @@ public function testProcessConfigurationWithDefaultConfiguration(): void
                 'twig' => [
                     'enabled' => class_exists(TwigBundle::class),
                 ],
+                'cache' => [
+                    'enabled' => interface_exists(CacheInterface::class),
```
Jean85, line 51: This needs to be added to the `full` configuration fixture.

Jean85 (changes requested):

CHANGELOG.md
ste93cry: It's a duplicate entry (fix has been released in `4.0.1`), probably the result of a bad merge between `master` and `develop`

tests/DependencyInjection/ConfigurationTest.php
ste93cry: Does it makes sense? Unlike the other instrumentations, this one does not register any service by default so there is nothing to test, apart from the compiler pass

Jean85: Services get still decorated; so it makes sense to test that somehow in end2end.

ste93cry: The decoration is tested in the `CompilerPassTest` class, it's kinda of an integration test as it really uses the `ContainerBuilder` to execute all steps 😉 The only thing that is mocked is the service definition of the abstract services

Jean85 (approved):

Jean85:
Build should be fixed by #480

VincentLanglet:
Hi @ste93cry  Could some documentation of this option be added in the Readme ?

ste93cry:
Please open an issue in [`getsentry/sentry-docs`](https://github.com/getsentry/sentry-docs), documentation indeed needs some updating

VincentLanglet:
Done https://github.com/getsentry/sentry-docs/issues/3463

PR #480, opened 2021-04-09
Force installing the same version for all Symfony components during the CI

ste93cry:
While trying to understand why some end-to-end tests recently started to fail during the CI due to deprecation notices coming from some Symfony components, I found out that components from different versions are installed at the same time. For example, let's suppose that we want to test the compatibility of our lib with Symfony `3.4`: what the current GitHub Action workflow does is replacing the version of all entries that start with `symfony/` in the `composer.json` with `3.4.x`.

https://github.com/getsentry/sentry-symfony/blob/43cd505db0a9e370326226467a8a61c7c7deb52f/.github/workflows/tests.yaml#L59

However, this does not guarante at all that the dependencies of those packages are installed at that version too. In fact, Symfony components often declare their compatibility with older or newer versions as well, but there is no guarantee that no deprecation is thrown if you mix two major versions. With this PR, I'm installing Flex globally and then using the `SYMFONY_REQUIRE` environment variable we force the same version for all components. I also discovered that apparently the `4.0` version of the SDK was uninstallable on Symfony `3.4` in certain cases because we were requiring `symfony/psr-http-message:^2.0` which needs some components from Symfony `4.x` instead, so I widened the Composer constraint to allow installing an older version of the package: this effectively allows installing our lib on a project that uses only components of the `3.4` version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "symfony/event-dispatcher": "^3.4.44||^4.4.20||^5.0.11",
         "symfony/http-kernel": "^3.4.44||^4.4.20||^5.0.11",
         "symfony/polyfill-php80": "^1.22",
-        "symfony/psr-http-message-bridge": "^2.0",
+        "symfony/psr-http-message-bridge": "^1.2||^2.0",
         "symfony/security-core": "^3.4.44||^4.4.20||^5.0.11"
     },
     "require-dev": {
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,6 +12,8 @@ jobs:
   Tests:
     runs-on: ubuntu-latest
     continue-on-error: ${{ matrix.experimental == true }}
+    env:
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -29,7 +29,7 @@
         "symfony/event-dispatcher": "^3.4.44||^4.4.20||^5.0.11",
         "symfony/http-kernel": "^3.4.44||^4.4.20||^5.0.11",
         "symfony/polyfill-php80": "^1.22",
-        "symfony/psr-http-message-bridge": "^2.0",
+        "symfony/psr-http-message-bridge": "^1.2||^2.0",
```
Jean85, line 32: Should we add a changelog entry due to this fix?

Jean85 (changes requested):

composer.json
ste93cry: Done 😃

Jean85 (approved):

PR #489, opened 2021-04-26
Fix deprecations triggered by Symfony 5.3

derrabus:
This PR fixes deprecation warnings from the upcoming Symfony 5.3 release:

* The "master request" has been renamed to "main request".
* Implementations of `UserInterface` must provide a `getUserIdentifier()` method now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix the conditions to automatically enable the cache instrumentation when possible (#487)
+- Fix deprecations triggered by Symfony 5.3 (#489)
 
 ## 4.1.0 (2021-04-19)
 
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -181,17 +181,17 @@ parameters:
 			path: tests/EventListener/RequestListenerTest.php
 
 		-
-			message: "#^Call to method isMasterRequest\\(\\) on an unknown class Symfony\\\\Component\\\\HttpKernel\\\\Event\\\\GetResponseEvent\\.$#"
-			count: 1
-			path: tests/EventListener/SubRequestListenerTest.php
-
-		-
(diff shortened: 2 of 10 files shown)


src/EventListener/EventForwardCompatibilityTrait.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Sentry\SentryBundle\EventListener;
+
+use Symfony\Component\HttpKernel\Event\KernelEvent;
+
+/**
+ * Provides forward compatibility with newer Symfony versions.
+ *
+ * @internal
+ */
+trait EventForwardCompatibilityTrait
```
ste93cry, line 14: I suggest to rename this class to `KernelEventForwardCompatibilityTrait` so that the naming is aligned to the one for which the compatibility trait is needed

src/EventListener/EventForwardCompatibilityTrait.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    private function isMainRequest(KernelEvent $event): bool
```
ste93cry, line 16: If we make this method `protected` we can use it in `AbstractTracingRequestListener` instead of in the implementations of that class

ste93cry (changes requested):
Lovely work. Can you :pray: add an entry to the CHANGELOG following the pattern of the existing ones?

src/EventListener/EventForwardCompatibilityTrait.php
derrabus: Good idea!

derrabus:
Done!

I noticed that some of my changes could also be applied to the 3.5.x branch. Is that branch still open for this kind of fixes? Would it make sense to fix that branch first then?

ste93cry:
>Is that branch still open for this kind of fixes? Would it make sense to fix that branch first then?

Yes it is, if you have time to contribute also to that version it would be awesome. Unfortunately we cannot merge the two branches because the code is foundamentally different and so we have to backport each commit/PR manually

ste93cry (approved):

derrabus:
Done in #490.

Issue #482, opened 2021-04-19
Custom RepresentationSerializer


olsavmic:
Hi,
with version 4, the configuration has been rewritten and https://github.com/getsentry/sentry-symfony/pull/255 no longer applies.

In order to rewrite the default Sentry RepresenationSerializer, the factory definition must be rewritten as shown here: 

```php
$sentryClientDefinition = $containerBuilder
	->getDefinition('sentry.client')
	->getFactory();

assert(is_array($sentryClientDefinition));
$factoryDefinition = $sentryClientDefinition[0];
assert($factoryDefinition instanceof Definition);

$factoryDefinition->addMethodCall('setRepresentationSerializer', [
	$containerBuilder->getDefinition(SentryReprSerializer::class),
]);
```

Given the fact that the RepresentationSerializer provides some configuration options like maxDepth and serializeAllObjects, I'd say it'd be worth it to simplify such configuration.

Or is there any other way to configure what the serializer that I'm missing?

Thank you!

Jean85:
`maxDepth` and `serializeAllObjects` are surely injectable only into that class, but depending on what do you want to achieve, you can implement the `Sentry\Serializer\SerializableInterface` on your objects to help with that without having to change this bundle's configuration.

Jean85:
Also, PR are welcome to fix the doc.

olsavmic:
Well I can improve the docs but I would say that such compiler pass is rather a hack than proper solution.

I know of SerializableInterface and class_serializers but there are use cases when you need to modify "global" behaviour.

I'd prefer to add the representation serializer (and perhaps even the serializer) as optional configuration options.

I can prepare a PR if such modification is a welcome change

olsavmic:
I'm not sure if we understand each other, I created a working PoC here: https://github.com/getsentry/sentry-symfony/pull/491

PR #491, opened 2021-04-28
Allow to pass custom serializer and representation_serializer in configuration

olsavmic:
Resolves https://github.com/getsentry/sentry-symfony/issues/482

> Well I can improve the docs but I would say that such compiler pass is rather a hack than proper solution.
> The problem with the compiler pass is that it's a modification of the internals and not the public API so it's a subject to change in future releases without any warning.
>I know of SerializableInterface and class_serializers but there are use cases when you need to modify "global" behaviour.
> I'd prefer to add the representation serializer (and perhaps even the serializer) as optional configuration options.
>I can prepare a PR if such modification is a welcome change.

### DIFF
--- a/tests/DependencyInjection/SentryExtensionTest.php
+++ b/tests/DependencyInjection/SentryExtensionTest.php
@@ -361,6 +361,24 @@ abstract class SentryExtensionTest extends TestCase
         $this->assertFalse($container->hasDefinition(TwigTracingExtension::class));
     }
 
+    public function testRepresentationSerializerOption(): void
+    {
+        $container = $this->createContainerFromFixture('custom_representation_serializer');
+        $factory = $container->findDefinition('sentry.client')->getFactory();
+
+        $methodCalls = $factory[0]->getMethodCalls();
+        $this->assertDefinitionMethodCallAt($methodCalls[4], 'setRepresentationSerializer', [new Reference('custom_representation_serializer')]);
+    }
+
+    public function testSerializerOption(): void
+    {
+        $container = $this->createContainerFromFixture('custom_serializer');
+        $factory = $container->findDefinition('sentry.client')->getFactory();
+
+        $methodCalls = $factory[0]->getMethodCalls();
+        $this->assertDefinitionMethodCallAt($methodCalls[3], 'setSerializer', [new Reference('custom_serializer')]);
+    }
+
     private function createContainerFromFixture(string $fixtureFile): ContainerBuilder
     {
         $container = new ContainerBuilder(new EnvPlaceholderParameterBag([
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Add option to provide custom `Serializer` and `RepresentationSerializer` via configuration (#491)
 
 - Make the transport factory configurable in the bundle's config (#504)
 
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Keep in mind that leaving the `dsn` value empty (or undeclared) in other environ
 ```yaml
 sentry:
     dsn: "https://public:secret@sentry.example.com/1"
-    messenger: 
+    messenger:
         enabled: true # flushes Sentry messages at the end of each message handling
         capture_soft_fails: true # captures exceptions marked for retry too
     options:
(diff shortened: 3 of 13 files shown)


olsavmic:
Custom serializer can then be registered as a service as usually

```yaml
App\Logger\SentryReprSerializer:
    arguments:
        $options: '@sentry.client.options'
```

olsavmic:
@Jean85 @ste93cry I did not receive any comment on this issue. Can I ask for a review of this PR or a comment on the related issue so we can proceed or close the issue? 

Thanks a lot!

Jean85 (changes requested):
Code looks fine for what I can see. It would require a changelog entry though.
Also, I'll retarget do `develop` since it's a new feature.

@ste93cry WDYT? Is this a valuable feature in your opinion?

tests/DependencyInjection/SentryExtensionTest.php
```diff
@@ -362,6 +362,54 @@ public function testTwigTracingExtensionIsRemovedWhenTwigTracingIsDisabled(): vo
...
+
+    public function testCustomSerializer(): void
+    {
+        $expectedSerializerReference = new Reference('custom_serializer');
+
+        $container = $this->createContainerFromFixture('custom_serializer');
+        $clientBuilderDefinition = $container->findDefinition('sentry.client')->getFactory()[0];
+        $clientBuilderMethodCalls = $clientBuilderDefinition->getMethodCalls();
+
+        $setSerializerCalls = array_filter(
```
ste93cry, line 397: Can you please use `assertDefinitionMethodCallAt` rather than this code? Your version is more future-proof, but I prefer to keep the consistency with the other tests until we decide to refactor the assertion helper

tests/DependencyInjection/SentryExtensionTest.php
```diff
@@ -362,6 +362,54 @@ public function testTwigTracingExtensionIsRemovedWhenTwigTracingIsDisabled(): vo
...
 
+    public function testRepresentationSerializer(): void
+    {
+        $expectedSerializerReference = new Reference('custom_representation_serializer');
+
+        $container = $this->createContainerFromFixture('custom_representation_serializer');
+        $clientBuilderDefinition = $container->findDefinition('sentry.client')->getFactory()[0];
+        $clientBuilderMethodCalls = $clientBuilderDefinition->getMethodCalls();
+
+        $setSerializerCalls = array_filter(
```
ste93cry, line 373: Can you please use `assertDefinitionMethodCallAt` rather than this code? Your version is more future-proof, but I prefer to keep the consistency with the other tests until we decide to refactor the assertion helper

tests/DependencyInjection/SentryExtensionTest.php
```diff
@@ -362,6 +362,54 @@ public function testTwigTracingExtensionIsRemovedWhenTwigTracingIsDisabled(): vo
...
+        $this->assertCount(1, $setSerializerCalls);
+
+        // Retrieve first argument passed to setRepresentationSerializer method call
+        $setSerializerArgument = $setSerializerCalls[array_key_first($setSerializerCalls)][1][0];
+        \assert($setSerializerArgument instanceof Reference);
+
+        $this->assertEquals($setSerializerArgument, $expectedSerializerReference);
+    }
+
+    public function testCustomSerializer(): void
```
ste93cry, line 389: Please rename this method to `testSerializerOption`

tests/DependencyInjection/SentryExtensionTest.php
```diff
@@ -362,6 +362,54 @@ public function testTwigTracingExtensionIsRemovedWhenTwigTracingIsDisabled(): vo
         $this->assertFalse($container->hasDefinition(TwigTracingExtension::class));
     }
 
+    public function testRepresentationSerializer(): void
```
ste93cry, line 365: Please rename this method to `testRepresentationSerializerOption`

ste93cry (changes requested):
>Is this a valuable feature in your opinion?

Yes, it is. In the future though I would like to see both serializers gone and replaced with either a refactored version or a third party lib so that we don't have to maintain the mess they are right now

ste93cry:
Stricly speaking about the current PR's code I have nothing more to say besides great work, however I'm now wondering if this is the best way to tackle this problem. The fact is that the serializer requires the options of the client whose service is not defined until it gets created by the `SentryExtension` during the compilation phase of the container. In turn, when I coded the service of the options I made it `private` because my though was (and is) that to access them you should pass from the client: the options is not something meant to be injected as a normal service because it belongs to a specific client instance. Maybe, instead of allowing to set the serializer instance, we should instead use a factory that given the options returns an instance of the serializer, forcibly making people aware that each client must have its own serializer instance (which usually doesn't happen)?

PR #492, opened 2021-04-28
Add bus, receiver and message name to extra's

airoude:
Hello,

With this PR we can send extra information about the worker like busName, receiver, and the actual message object name.
In the future (when I have more time) we can add configuration for custom stamps to be sent along.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog
 
 ## Unreleased
-- Attach bus, receiver and message class name as extra's
 
+- Log the bus name, receiver name and message class name as event tags when using Symfony Messenger (#492)
 - Make the transport factory configurable in the bundle's config (#504)
 - Add the `sentry_trace_meta()` Twig function to print the `sentry-trace` HTML meta tag (#510)
 - Make the list of commands for which distributed tracing is active configurable (#515)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Attach bus, receiver and message class name as extra's
 
 - Make the transport factory configurable in the bundle's config (#504)
 - Add the `sentry_trace_meta()` Twig function to print the `sentry-trace` HTML meta tag (#510)
--- a/src/EventListener/MessengerListener.php
+++ b/src/EventListener/MessengerListener.php
@@ -5,9 +5,15 @@ declare(strict_types=1);
 namespace Sentry\SentryBundle\EventListener;
 
 use Sentry\State\HubInterface;
+use Sentry\State\Scope;
+use Symfony\Component\Messenger\Envelope;
 use Symfony\Component\Messenger\Event\WorkerMessageFailedEvent;
 use Symfony\Component\Messenger\Event\WorkerMessageHandledEvent;
 use Symfony\Component\Messenger\Exception\HandlerFailedException;
+use Symfony\Component\Messenger\Stamp\BusNameStamp;
+use Throwable;
+
+use function get_class;
 
 final class MessengerListener
 {
@@ -49,10 +55,18 @@ final class MessengerListener
 
         if ($error instanceof HandlerFailedException) {
             foreach ($error->getNestedExceptions() as $nestedException) {
-                $this->hub->captureException($nestedException);
+                $this->captureExceptionWithScope(
+                    $nestedException,
+                    $event->getEnvelope(),
+                    $event->getReceiverName()
+                );
             }
         } else {
-            $this->hub->captureException($error);
+            $this->captureExceptionWithScope(
+                $error,
+                $event->getEnvelope(),
+                $event->getReceiverName()
+            );
         }
 
         $this->flushClient();
@@ -77,5 +91,25 @@ final class MessengerListener
         if (null !== $client) {
             $client->flush();
         }
+    }
+
+    private function captureExceptionWithScope(
+        Throwable $exception,
+        Envelope $envelope,
+        string $receiverName
+    ): void {
+        $this->hub->withScope(function(Scope $scope) use ($receiverName, $envelope) {
+            if ($messageBusStamp = $envelope->last(BusNameStamp::class)) {
+                assert($messageBusStamp instanceof BusNameStamp);
+                $scope->setTag('messenger.message_bus', $messageBusStamp->getBusName());
+            }
+
+            $scope->setTags([
+                'messenger.receiver_name' => $receiverName,
+                'messenger.message_class' => get_class($envelope->getMessage())
+            ]);
+        });
+
+        $this->hub->captureException($exception);
     }
 }
--- a/src/EventListener/MessengerListener.php
+++ b/src/EventListener/MessengerListener.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Sentry\SentryBundle\EventListener;
 
+use Sentry\Event;
+use Sentry\EventHint;
+use Sentry\Severity;
 use Sentry\State\HubInterface;
 use Sentry\State\Scope;
 use Symfony\Component\Messenger\Envelope;
@@ -11,9 +14,6 @@ use Symfony\Component\Messenger\Event\WorkerMessageFailedEvent;
 use Symfony\Component\Messenger\Event\WorkerMessageHandledEvent;
 use Symfony\Component\Messenger\Exception\HandlerFailedException;
 use Symfony\Component\Messenger\Stamp\BusNameStamp;
-use Throwable;
-
-use function get_class;
 
 final class MessengerListener
 {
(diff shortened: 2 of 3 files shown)


tests/EventListener/MessengerListenerTest.php
```diff
@@ -55,6 +56,9 @@ public function testHandleWorkerMessageFailedEvent(array $exceptions, WorkerMess
             ->method('getClient')
             ->willReturn($this->client);
 
+        $this->hub->expects($this->exactly(\count($exceptions)))
+            ->method('withScope');
```
ste93cry, line 60: Since `withScope` is basically a wrapper that just calls the given callback by providing it a new scope, you can simulate this behaviour using something like the code below (untested) and asserting that the data is correctly set

```php
->method('withScope')
->willReturnCallback(static function (callable $callback) use ($scope): void {
    $callback($scope);

    $this->assertSame(...,);
});
```

tests/EventListener/MessengerListenerTest.php
```diff
@@ -155,8 +163,14 @@ public function testHandleWorkerMessageHandledEvent(): void
         $this->client->expects($this->once())
             ->method('flush');
 
+        $event = new WorkerMessageHandledEvent(
```
ste93cry, line 166: Instead of editing this test case I suggest to add a new one. The reason is that we should ensure that the code works both when the `stamp` is present and when it isn't. In this specific case, unlike the other reviewed places, we should also add the data provider

ste93cry (changes requested):
Can you please also add a CHANGELOG entry following the pattern of the existing entries?

src/EventListener/MessengerListener.php
```diff
@@ -78,4 +92,24 @@ private function flushClient(): void
...
+                $scope->setTag('messenger.message_bus', $messageBusStamp->getBusName());
+            }
+
+            $scope->setTags([
+                'messenger.receiver_name' => $receiverName,
+                'messenger.message_class' => get_class($envelope->getMessage())
+            ]);
+        });
+
+        $this->hub->captureException($exception);
```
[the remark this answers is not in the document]
airoude: I couldn't find any examples where ```withScope```  was used in combination with ```captureException```. If I put it inside the ```withScope``` it doesn't get invoked. Any tips?

src/EventListener/MessengerListener.php
```diff
@@ -78,4 +92,24 @@ private function flushClient(): void
...
         }
     }
+
+    private function captureExceptionWithScope(
+        Throwable $exception,
+        Envelope $envelope,
+        string $receiverName
+    ): void {
+        $this->hub->withScope(function(Scope $scope) use ($receiverName, $envelope) {
+            if ($messageBusStamp = $envelope->last(BusNameStamp::class)) {
```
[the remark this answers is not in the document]
airoude: Hm, why? This already checks on null.

src/EventListener/MessengerListener.php
```diff
@@ -78,4 +92,24 @@ private function flushClient(): void
...
+                assert($messageBusStamp instanceof BusNameStamp);
```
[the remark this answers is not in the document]
airoude: The assert gives type-hinting and it validates.

src/EventListener/MessengerListener.php
```diff
@@ -78,4 +92,24 @@ private function flushClient(): void
...
+                $scope->setTag('messenger.message_bus', $messageBusStamp->getBusName());
```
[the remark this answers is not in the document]
airoude: Yes valid point.

src/EventListener/MessengerListener.php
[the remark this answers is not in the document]
ste93cry: I prefer to be explicit and I think it makes more clear the intent of what is happening here for the reader

tests/EventListener/MessengerListenerTest.php
airoude: Where is the ```$scope``` coming from?

src/EventListener/MessengerListener.php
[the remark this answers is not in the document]
ste93cry: Quoting from the official PHP documentation, their use should be mainly reserved as a debugging feature. In this context, I don't think it's the case: we trust the vendor to return an instance of the object expected, otherwise there would be a bug there. I think that an annotation is more than enough, even if we lose the so called "validation"...

src/EventListener/MessengerListener.php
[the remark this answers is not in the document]
airoude: This pattern is broadly used in Symfony, Doctrine, and Composer. But I will make it explicit.

tests/EventListener/MessengerListenerTest.php
ste93cry: Oh sorry, I partially copy-pasted the code from some other place of the codebase. In this case, `$scope` can be instantiated outside the lambda (and consequently the assertions too can be moved to the outside if you prefer) or you can initialize the variable inside the lambda and avoid passing it from the outside. It's your call

airoude: Thanks!

tests/EventListener/MessengerListenerTest.php
```diff
@@ -89,6 +91,72 @@ public function handleWorkerMessageFailedEventDataProvider(): \Generator
...
+                $callback($scope);
+
+                // $scope has no tags getters, should make assertions against this
+                // 'messenger.receiver_name => 'receiver',
+                // 'messenger.message_class' => 'stdClass',
+                // 'messenger.message_bus' => 'commandBus',
+            });
+
+        $this->hub->expects($this->exactly(\count($exceptions)))
+            ->method('captureEvent');
```
airoude, line 118: Idk why this one fails. @ste93cry could you help?

ste93cry: It's failing because you are not mocking the `withScope()` method, therefore the callback function is never invoked

airoude: Isn't it mocked on L106?

ste93cry: Yes, but this line is referring to the `testHandleWorkerMessageFailedEventWithStamps()` test method, while the failing one is `testHandleWorkerMessageFailedEvent()`

airoude: Oh jeez.. a long week. Thanks for the catch!

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Attach bus, receiver and message class name as extra's
```
ste93cry, line 4: Please add a reference to this PR

src/EventListener/MessengerListener.php
```diff
@@ -78,4 +91,34 @@ private function flushClient(): void
...
     }
+
+    private function captureException(
+        \Throwable $exception,
+        Envelope $envelope,
+        string $receiverName
+    ): void {
+        $event = Event::createEvent();
+        $event->setLevel(Severity::error());
+        $exceptionHint = EventHint::fromArray([
```
ste93cry, line 102: I don't understand why you are manually creating the event here, `captureException()` was fine as it was and using `captureEvent()` does not have any benefit here apart from making the code unnecessarily more complicated

tests/EventListener/MessengerListenerTest.php
```diff
@@ -45,11 +47,20 @@ public function testHandleWorkerMessageFailedEvent(array $exceptions, WorkerMess
...
-                return [$exception];
-            }, $exceptions));
+            ->method('withScope')
+            ->willReturnCallback(function (callable $callback) use ($scope): void {
+                $callback($scope);
+
+                // $scope has no tags getters, should make assertions against this
+                // 'messenger.receiver_name => 'receiver',
+                // 'messenger.message_class' => 'stdClass',
+                // 'messenger.message_bus' => 'commandBus',
```
ste93cry, line 59: While it's true that the scope has no getters, you should call `$scope->applyToEvent(Event::createEvent())` and then check that the data is set on the event itself

ste93cry (changes requested):
I see there is still some commented code [here and there](https://github.com/getsentry/sentry-symfony/pull/492/files#diff-188d24189ff68e40532a471f178a0824ab89645b1c0bb35e3e3e8dcbc30d9858R120), and I don't understand why you switched from `captureException()` to `captureEvent()`. I mentioned the latter in https://github.com/getsentry/sentry-symfony/pull/492#discussion_r622498062 just for the sake of the example, I didn't mean you to change it

ste93cry:
I took the time to finish this PR as I would really love to see it merged. Hopefully you don't mind I rebased the PR and pushed changes to your code. It would be great if you could test it out before I give the final approval to be sure everything works as expected

Jean85 (approved):

ste93cry (approved):
Since all is silent, LGTM 🚢 Thanks @airoude for the work you did

PR #495, opened 2021-05-04
Fix fatal error when the `SERVER_PROTOCOL` header is missing

### LINKED ISSUE #494: Crash when using php-fpm fcgi api without HTTP server
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
```
sentry/sdk 3.1.0
sentry/sentry 3.2.1
sentry/sentry-symfony 4.1.0
```

### Steps to Reproduce

We have a healthcheck script that monitors the status of php-fpm like so:

```
REQUEST_METHOD=GET REQUEST_URI=/_health SCRIPT_FILENAME=index.php cgi-fcgi -bind -connect 127.0.0.1:9000
```
In this case the call to `$request->getProtocolVersion()` returns `null` which results in an error in `TracingRequestListener::getHttpFlavor`

Error:
```
str_starts_with(): Argument #1 ($haystack) must be of type string, null given
```

### Expected Result

Eventhough this is a weird edge case that probably won't occur often maybe fall back to some empty value?

### Actual Result

I get an error (in sentry though :+1:)

Error:
```
str_starts_with(): Argument #1 ($haystack) must be of type string, null given
```

ste93cry:
As per title, fixes #494. It should not happen since the `SERVER_PROTOCOL` header  must be present according to [RFC 3875 (GCI spec)](https://tools.ietf.org/html/rfc3875#section-4.1.16), but in cases when it's mistakenly missing better be safe than sorry

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -16,6 +16,11 @@ parameters:
 			path: src/EventListener/ErrorListener.php
 
 		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\EventListener\\\\TracingRequestListener\\:\\:getHttpFlavor\\(\\) never returns null so it can be removed from the return typehint\\.$#"
+			count: 1
+			path: src/EventListener/TracingRequestListener.php
+
+		-
 			message: "#^Call to an undefined method Doctrine\\\\DBAL\\\\Driver\\|Sentry\\\\SentryBundle\\\\Tracing\\\\Doctrine\\\\DBAL\\\\Compatibility\\\\ExceptionConverterDriverInterface\\:\\:connect\\(\\)\\.$#"
 			count: 1
 			path: src/Tracing/Doctrine/DBAL/TracingDriver.php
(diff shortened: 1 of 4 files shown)


phpstan-baseline.neon
```diff
@@ -15,6 +15,11 @@ parameters:
 			count: 1
 			path: src/EventListener/ErrorListener.php
 
+		-
+			message: "#^Method Sentry\\\\SentryBundle\\\\EventListener\\\\TracingRequestListener\\:\\:getHttpFlavor\\(\\) never returns null so it can be removed from the return typehint\\.$#"
```
Jean85, line 19: We could suggest this fix upstream.

Jean85 (approved):

phpstan-baseline.neon
ste93cry: Absolutely, but until this is fixed we have to keep this 😢

Issue #497, opened 2021-05-05
Injecting own Spool into SpoolTransport 


graferowy:
Hi!

How can I make sentry to use SpoolTransport (taken from one of repository's branches- https://github.com/getsentry/sentry-php/blob/e7fb87f840ab026e9ee9444831b6c7c571f06ed0/src/Transport/SpoolTransport.php). I copied this Transport into my project files (and adjusted it to the current TransportInterface) and how can I now set it as a default transport with spools defined by me?

Using
sentry-symfony 4.1
symfony 5.1

graferowy:
Thank you for your answer. What I end up doing was to get the factory of `sentry.client` definition in my compiler pass. Then I added `setTransportFactory` method call with my spool transport factory. Everything seems to be working fine so I hope it's a valid solution.

`
$transportFactory = $container->getDefinition(SpoolTransportFactory::class);
$sentryClient = $container->findDefinition('sentry.client');
$factory = $sentryClient->getFactory();

if (isset($factory[0]) && $factory[0] instanceof Definition) {
     $factory[0]->addMethodCall('setTransportFactory', [$transportFactory]);
}
`

ste93cry:
Yes, that's another solution indeed. Anyway, maybe in the future we could make some things like setting the transport factory configurable so that you don't have to write hacky compiler passes

ste93cry:
I'm closing this because I think it's solved. Btw, in the upcoming `4.2` version it will be possible to configure the factory in the bundle's config

PR #504, opened 2021-05-16
Make the transport factory configurable in the bundle's config

ste93cry:
While answering issue #497 I realized that making the transport factory configurable in the bundle's config so that there is no need to use a compiler pass to modify the service definition of the SDK client would make things much more easy for the users. Change itself is pretty small, it just adds a new config node that can be set to the name of the service to use in place of the default one 😄

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -8,6 +8,7 @@ use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use Jean85\PrettyVersions;
 use Sentry\Options;
 use Sentry\SentryBundle\ErrorTypesParser;
+use Sentry\Transport\TransportFactoryInterface;
 use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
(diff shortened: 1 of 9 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -35,6 +36,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->info('If this value is not provided, the SDK will try to read it from the SENTRY_DSN environment variable. If that variable also does not exist, the SDK will just not send any events.')
                 ->end()
                 ->booleanNode('register_error_listener')->defaultTrue()->end()
+                ->scalarNode('transport_factory')
+                    ->info('The service ID of the transport factory used by the default SDK client.')
+                    ->defaultValue(TransportFactoryInterface::class)
```
Jean85, line 41: I'm not sure about this default. This would mean that declaring a service with that classname would automatically override the transport. Is this what we want?

ste93cry: That's precisely what this feature aims for: a simple way to set which service representing the transport factory is used. Also note that this default value is the same [as before](https://github.com/getsentry/sentry-symfony/pull/504/files#diff-321b6d04474e929d9287516b7a39b25b5c499de506dd0bda2f7940d7a148a389L131), I only moved where it was set from the DI extension class to the config

Jean85 (approved):

PR #506, opened 2021-05-18
Cache contracts is a required dependency (not a dev-dependency)

matthew-gill:
I installed this locally and found that I got exceptions for missing classes: `CacheInterface` (and a couple more that I don't remember, sorry).

Turns out that the `symfony/cache-contracts` https://symfony.com/doc/current/components/contracts.html is indeed required as a dependency, and not a dev-dependency.


Jean85:
Can you report to us the stack traces that lead to those errors? It's an optional dependency due to tracing, we're not strictly requiring it.

Maybe it's a regression after #501?

ste93cry:
It cannot be a regression because that fix has not been released yet 😁 Btw, I believe that the referenced PR solved this exact issue, we're gonna release the fix in the next days so let's wait for it and see if it solves the problem

matthew-gill:
Unfortunately I wasn't able to pull a stack trace as it was a fatal error.

It came from `src/Tracing/Cache/TraceableCacheAdapter.php`, though.

Jean85:
@matthew-gill #501 got released as 4.1.2, can you check if that fixes your issue?

ste93cry (approved):

ste93cry:
I just found out trying to fix #511 that the `symfony/cache-contracts` package is not optional because our cache implementation used to support the distributed tracing feature depends on it, but on Symfony `3.4` the Cache component doesn't.

PR #510, opened 2021-05-19
Add the `sentry_trace_meta()` Twig function to print the `sentry-trace` HTML meta tag

### LINKED ISSUE #458: Utility for printing the meta tag for tracing
As reported in https://github.com/getsentry/sentry-php/issues/1192#issuecomment-789709631, to link a PHP tracing transaction to the frontend where you may be using the Sentry JS SDK, 

> You need to connect the PHP traces to the JS SDK by setting a meta tag on your page the JS SDK can read.
> ```
> <meta name="sentry-trace" content="49879feb76c84743ba5034bd2d3f1ca3-7cb5535c930d4666-1"/>
> ```
> Example from Laravel: https://docs.sentry.io/platforms/php/guides/laravel/performance/#connect-your-frontend.
> 
> This is how that header is built: https://github.com/getsentry/sentry-laravel/blob/be9658a57d554f8bfe587facc1492a45c6bd0e1c/src/Sentry/Laravel/Integration.php#L175-L187.

We could provide something similar here in the Symfony bundle; probably a Twig partial which uses a Twig function to retrieve the current transaction id.

ste93cry:
As per title, fixes #458 by adding the `sentry_trace_meta()` Twig function. Unlike the Laravel's SDK, I decided to always print the HTML meta tag regardless of whether a transaction/span is active because I like to be consistent in the expected output and because there should be no issue in having a blank string as its content

### DIFF
--- a/src/Twig/SentryExtension.php
+++ b/src/Twig/SentryExtension.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\SentryBundle\Twig;
+
+use Sentry\State\HubInterface;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+final class SentryExtension extends AbstractExtension
+{
+    /**
+     * @var HubInterface The current hub
+     */
+    private $hub;
+
+    /**
+     * @param HubInterface $hub The current hub
+     */
+    public function __construct(HubInterface $hub)
+    {
+        $this->hub = $hub;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('sentry_trace_meta', \Closure::fromCallable([$this, 'getTraceMeta']), ['is_safe' => ['html']]),
+        ];
+    }
+
+    /**
+     * Returns an HTML meta tag named `sentry-trace`.
+     */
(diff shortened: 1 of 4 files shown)


src/Twig/SentryExtension.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Returns an HTML meta tag named `sentry-trace`.
+     */
+    private function getTraceMeta(): string
+    {
+        $span = $this->hub->getSpan();
+
+        return sprintf('<meta name="sentry-trace" content="%s" />', null !== $span ? $span->toTraceparent() : '');
```
Jean85, line 43: I know it's a nitpick, but I had to double take because it seemed "transparent" :sweat_smile: 
```suggestion
        return sprintf('<meta name="sentry-trace" content="%s" />', null !== $span ? $span->toTraceParent() : '');
```

Jean85 (changes requested):

src/Twig/SentryExtension.php
ste93cry: Although in PHP the function names are case-insensitive, I think it's an error to call them with another casing

Jean85: Ok, so I think we should fix this with a PR, it's not a BC.

Jean85 (approved):

PR #512, opened 2021-05-27
Rework DBAL compat layer

Jean85:
WIP, this includes #517 to check thoroughly for feasibility.

@ste93cry WDYT of this different approach?


tests/Tracing/Doctrine/DBAL/TracingDriverTest.php
```diff
@@ -264,3 +254,48 @@ interface StubExceptionConverterDriverInterface extends DriverInterface
     {
     }
 }
+
+class MockDriver implements DriverInterface
```
ste93cry, line 258: It's not clear to me why you added this class instead of mocking the interface like I was doing before. What's the benefit of doing this or the blocker that makes it impossible now to do it?

ste93cry (changes requested):
I think that by making `2.13` the minimum required version (they recently added back support for PHP 7.2) we could probably cleanup some other code because they added a forward compatibility layer

tests/Tracing/Doctrine/DBAL/TracingDriverTest.php
Jean85: Psalm was complaining. But it didn't fix the issue anyway, so we could rollback this.

Would bumping DBAL requirement create issues?

ste93cry: >Would bumping DBAL requirement create issues?

I don't think so, but the best thing is to try. We couldn't bump the minimum version before simply because there was no support for PHP 7.2, which is now back

Jean85: I've investigated it further. The only additional FC is https://github.com/doctrine/dbal/pull/4529, which is about `ResultStatement`, that we do not touch, so I don't see any advantages.

I'll try to push this PR forward, and on the code side it seems to work, but the static analysis is a complete mess.

tacman: >> there was no support for PHP 7.2, which is now back

But isn't 7.2 dead?

ste93cry: It's EOL, but we still support it mainly because Sentry must run also on legacy platforms, if possible

Jean85:
Closing in favor of #527

PR #513, opened 2021-05-28
Fix: Use json instead of yaml for a fenced code block

localheinz:
This pull request

- [x] uses `json` instead of `yaml` for a fenced code block


Jean85 (approved):
Thanks for catching this!

localheinz:
Thank you, @Jean85!

PR #515, opened 2021-05-28
Make the list of commands for which distributed tracing is active configurable

### LINKED ISSUE #493: Disable tracing for console commands
I just started using tracing, and so far so good. However, it also registers traces for console commands, while those should be excepted in my opinion, as they do not directly affect performance for the user.

ste93cry:
Fixes #493. I'm a bit unsure if the option should be a list of commands or a regex string. The pros of the latter solution would be that it would be easier to exclude a lot of commands if their name shares the same pattern, but working with a list is easier than working with a regexp


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Make the transport factory configurable in the bundle's config (#504)
 - Add the `sentry_trace_meta()` Twig function to print the `sentry-trace` HTML meta tag (#510)
+- Make the list of commands for which distributed tracing is active configurable (#515)
```
Jean85, line 7: Should we note that we're excluding the consume command by default?

ste93cry: I don't think it's necessary because I think in most cases (if not all) it's something not wanted

Jean85 (approved):

PR #516, opened 2021-05-28
Fix missing require of the `symfony/cache-contracts` package

### LINKED ISSUE #511: Installation of sentry/sentry-symfony package in version 4.1 with symfony/symfony 3.4.x version breaks appliciation due to missing cache contract
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
`"sentry/sentry-symfony": "^4.1",`

### Steps to Reproduce

Installed the above mentioned package with

`composer require sentry/sentry-symfony`

### Expected Result

Application runs as before with an activated sentry integration.

### Actual Result
Application is broken:

Bildschirmfoto 2021-05-27 um 12 28 09

due to a missing symfony cache interface. The cache Interface used here:

https://github.com/getsentry/sentry-symfony/blob/master/src/Tracing/Cache/TraceableCacheAdapter.php#L11

but the compose.json file is not defining the `symfony/cache` or `symfony/cache-contracts` as required package.

ste93cry:
Fixes #511: the `symfony/cache-contracts` package is not optional because our cache implementation used to support the distributed tracing feature depends on it, but on Symfony `3.4` the [Cache component](https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Cache/composer.json) doesn't.


ste93cry:
Duplicate of #506

PR #525, opened 2021-06-16
Fix decoration of cache adapters inheriting parent service

### LINKED ISSUE #524: Cache tracing causes bug when using TagAwareCacheInterface
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?

`sentry.php.symfony` at `4.1.3`

### Steps to Reproduce

1. Have `sentry/sentry-symfony` installed at `4.1.3` and make sure to enable cache tracing.
`config/packages/sentry.yaml`:
```yaml
sentry:
    dsn: '%env(SENTRY_DSN)%'
    tracing:
        cache:
            enabled: true
```

2. Inject a cache pool `TagAwareCacheInterface $integrationsCache` into a service and use the `RedisTagAwareAdapter`:

```php
public function __construct(
    private int $cacheTtl,
    private TagAwareCacheInterface $integrationsCache,
    private NavisionCustomerService $navisionCustomerService,
) {}
```

`config/packages/cache.yaml`:
```yaml
framework:
    cache:
        prefix_seed: my_app
        app: cache.adapter.redis_tag_aware
        default_redis_provider: '%env(resolve:REDIS_DSN)%'
        pools:
            application.cache:
                adapter: cache.adapter.redis_tag_aware
                tags: true
            integrations.cache:
                adapter: cache.adapter.redis_tag_aware
                tags: true

```

3. Use the service in an HTTP request.

### Expected Result

Symfony should inject an instance `RedisTagAwareAdapter` into the service.

### Actual Result

An exception is thrown:
`
MyApp\Integrations\Navision\Customer\CachedNavisionCustomerService::__construct(): Argument #2 ($integrationsCache) must be of type Symfony\Contracts\Cache\TagAwareCacheInterface, Symfony\Component\Cache\Adapter\TraceableAdapter given, called in /vagrant/var/cache/dev/ContainerYqNNU93/getCachedNavisionCustomerServiceService.php on line 24
`

This exclusively happens when Sentry is installed and cache tracing is enabled. I feel this might be an issue in `Sentry\SentryBundle\DependencyInjection\Compiler\CacheTracingPass`.

ste93cry:
Fixes #524: when running a compiler pass during the `TYPE_BEFORE_OPTIMIZATION` phase, services inheriting a parent one have not been resolved yet and consequently their class is still set to `null`. This was causing the check to decorate the cache adapter according to whether they were implementations of the `TagAwareAdapterInterface` interface or not to fail, resulting in the decorator to always inherit the base `AdapterInterface` interface even when the adaptor should have been an instance of `TagAwareAdapterInterface`

### DIFF
--- a/src/DependencyInjection/Compiler/CacheTracingPass.php
+++ b/src/DependencyInjection/Compiler/CacheTracingPass.php
@@ -8,6 +8,7 @@ use Symfony\Component\Cache\Adapter\TagAwareAdapterInterface;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
 
 final class CacheTracingPass implements CompilerPassInterface
@@ -28,7 +29,13 @@ final class CacheTracingPass implements CompilerPassInterface
                 continue;
             }
 
-            if (is_subclass_of($cachePoolDefinition->getClass(), TagAwareAdapterInterface::class)) {
+            $definitionClass = $this->resolveDefinitionClass($container, $cachePoolDefinition);
+
+            if (null === $definitionClass) {
+                continue;
+            }
+
+            if (is_subclass_of($definitionClass, TagAwareAdapterInterface::class)) {
                 $traceableCachePoolDefinition = new ChildDefinition('sentry.tracing.traceable_tag_aware_cache_adapter');
             } else {
                 $traceableCachePoolDefinition = new ChildDefinition('sentry.tracing.traceable_cache_adapter');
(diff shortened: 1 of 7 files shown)


src/DependencyInjection/Compiler/CacheTracingPass.php
```diff
@@ -40,4 +47,16 @@ public function process(ContainerBuilder $container): void
...
         }
     }
+
+    private function resolveDefinitionClass(ContainerBuilder $container, Definition $definition): ?string
+    {
+        $class = $definition->getClass();
+
+        while ($definition instanceof ChildDefinition) {
+            $definition = $container->findDefinition($definition->getParent());
+            $class = $class ?: $definition->getClass();
```
Jean85, line 57: This loop doesn't seem correct to me... If `$class` is populated at first try, it basically ignores the previous line.

Maybe we can rewrite this in a more understandable way?

src/DependencyInjection/Compiler/CacheTracingPass.php
```diff
@@ -28,7 +29,13 @@ public function process(ContainerBuilder $container): void
                 continue;
             }
 
-            if (is_subclass_of($cachePoolDefinition->getClass(), TagAwareAdapterInterface::class)) {
+            $definitionClass = $this->resolveDefinitionClass($container, $cachePoolDefinition);
+
+            if (null === $definitionClass) {
+                continue;
```
Jean85, line 35: How is it possible that this line is never covered? Shouldn't the method "explode" if it can't find a FQCN for the service?

Jean85 (changes requested):

src/DependencyInjection/Compiler/CacheTracingPass.php
ste93cry: The `resolveDefinitionClass()` method can return `null`, and in this case instead of throwing an exception I decided to just skip decorating the service. The container compilation will do the rest anyway

Jean85 (approved):

PR #527, opened 2021-06-23
Bump DBAL support to 2.13+

### LINKED ISSUE #483: Deprecation warnings in 4.1.0
### Environment

How do you use Sentry?
self-hosted

Which SDK and version?

sentry-symfony 4.1.0
symfony 5.2.6

### Steps to Reproduce

Symfony reports a couple of deprecation warnings

```
1x: The "Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriver" class implements "Doctrine\DBAL\Driver\ExceptionConverterDriver" that is deprecated.

  1x: The "Sentry\State\HubAdapter::captureMessage()" method will require a new "EventHint|null $hint" argument in the next major version of its interface "Sentry\State\HubInterface", not defining it is deprecated.

  1x: The "Sentry\State\HubAdapter::captureException()" method will require a new "EventHint|null $hint" argument in the next major version of its interface "Sentry\State\HubInterface", not defining it is deprecated.

  1x: The "Sentry\State\HubAdapter::captureLastError()" method will require a new "EventHint|null $hint" argument in the next major version of its interface "Sentry\State\HubInterface", not defining it is deprecated.

```

### Expected Result

no warnings

Jean85:
~WIP~

This bumps the requirement for DBAL tracing to `^2.13 | ^3`; this in turn allows us to leverage the wider FC layer present in 2.13, reducing a lot our effort.

This is an alternative to #512, with a complete different approach as suggested in https://github.com/getsentry/sentry-symfony/pull/512#pullrequestreview-670685888

This should also fix #483 as a side effect.


src/Tracing/Doctrine/DBAL/TracingDriver.php
```diff
@@ -6,7 +6,7 @@
 
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Driver\API\ExceptionConverter;
-use Doctrine\DBAL\Driver as DriverInterface;
+use Doctrine\DBAL\Driver as DBALDriver;
```
ste93cry, line 9: Why not aliasing the interface? I did it mainly to keep consistency with our standard which is to have the `Interface` suffix, but if we don't want to alias this interface anymore then we should also not create aliases for the one that came below

src/Tracing/Doctrine/DBAL/TracingDriver.php
```diff
@@ -16,32 +16,30 @@
 use Sentry\State\HubInterface;
 
 /**
- * This is a simple implementation of the {@see DriverInterface} interface that
+ * This is a simple implementation of the {@see DBALDriver} interface that
  * decorates an existing driver to support distributed tracing capabilities.
- * This implementation IS and MUST be compatible with all versions of Doctrine
- * DBAL >= 2.10.
+ * This implementation IS cross-compatible with both DBAL 2 and 3 thanks to
```
ste93cry, line 21: This part of the documentation was more a warning that a simple note, and should stay the same because this class still implements things that got removed in DBAL `3.0`, but that are needed for implementations compatible with the `2.x` series

tests/Tracing/Doctrine/DBAL/TracingDriverTest.php
```diff
@@ -245,22 +246,30 @@ public function testConvertExceptionThrowsIfDoctrineDBALVersionIsAtLeast30(): vo
         $this->expectExceptionMessage('The Sentry\\SentryBundle\\Tracing\\Doctrine\\DBAL\\TracingDriver::convertException() method is not supported on Doctrine DBAL 3.0.');
 
         $driver = new TracingDriver($this->hub, $this->createMock(StubExceptionConverterDriverInterface::class));
+        /** @phpstan-ignore-next-line */
```
ste93cry, line 249: Maybe it makes sense to extract all errors to the PHPStan baseline to avoid cluttering the code with all these annotations?

tests/Tracing/Doctrine/DBAL/TracingDriverTest.php
```diff
@@ -245,22 +246,30 @@ public function testConvertExceptionThrowsIfDoctrineDBALVersionIsAtLeast30(): vo
...
         $driver->convertException('foo', $this->createMock(DriverExceptionInterface::class));
     }
 }
 
-if (interface_exists(VersionAwarePlatformDriverInterface::class)) {
-    interface StubVersionAwarePlatformDriverInterface extends DriverInterface, VersionAwarePlatformDriverInterface
-    {
+if (interface_exists(DriverInterface::class)) {
```
ste93cry, line 254: I don't get why you moved all the stubs declarations inside this conditional, which is always `true` as long as the DBAL is installed 🤔

ste93cry (changes requested):

src/DependencyInjection/Compiler/DbalTracingPass.php
```diff
@@ -39,6 +40,10 @@ public function process(ContainerBuilder $container): void
             return;
         }
 
+        if (!interface_exists(Driver::class)) {
+            throw new \RuntimeException(sprintf('The Doctrine "%s" interface is missing, DBAL connection cannot be instrumented; check that you have DBAL 2.13+ installed', Driver::class));
```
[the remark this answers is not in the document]
Jean85: You're right. This works by accident in CI due to the fact we `require-dev` `doctrine/dbal ^2.13`.
We should check for `\Doctrine\DBAL\Result`, which was introduced in 2.13.0 with https://github.com/doctrine/dbal/pull/4529/files#diff-ba0bc531c4d594368d24a7fea2d81f1aff2841ccf658fa2e4b51b9c822abb6d9R20

src/Tracing/Doctrine/DBAL/TracingDriver.php
Jean85: Reverted back.

tests/Tracing/Doctrine/DBAL/TracingDriverTest.php
Jean85: Because we have a CI job without DBAL installed; in that case, the test would explode due to missing classes.

src/DependencyInjection/Compiler/DbalTracingPass.php
```diff
@@ -92,4 +96,26 @@ private function getSetMiddlewaresMethodCallArguments(Definition $definition): a
...
         return [];
     }
+
+    /**
+     * @throws \RuntimeException
+     */
+    private function checkDbalVersion(): void
+    {
+        if (!interface_exists(Driver::class)) {
+            throw new \RuntimeException(sprintf('The Doctrine "%s" interface is missing, DBAL connection cannot be instrumented; check that you have DBAL installed', Driver::class));
```
ste93cry, line 106: I suggest to change the thrown exception to `LogicException` and the message to `Tracing support cannot be enabled as the Doctrine DBAL 2.13+ package is not installed. Try running "composer require doctrine/dbal:^2.13".` to be consistent with what Symfony does for its own components. Furthermore, we can refactorize this method to throw the exception just once at the end of the function when neither the `Doctrine\DBAL\Result` interface nor the class with the same name exist

ste93cry (changes requested):

tests/Tracing/Doctrine/DBAL/TracingDriverTest.php
ste93cry: How can it explode if the declaration of the stubs was already wrapped inside an `interface_exists()` statement?

Jean85: I think there was a version where that check was failing.

src/DependencyInjection/Compiler/DbalTracingPass.php
```diff
@@ -92,4 +96,24 @@ private function getSetMiddlewaresMethodCallArguments(Definition $definition): a
 
         return [];
     }
+
+    /**
+     * @throws \RuntimeException
```
ste93cry, line 101: This DocBlock is incorrect now that we've changed to throwing a `LogicException`. Anyway, I wonder if it makes sense to leave it because it provide little value considering that the method is not part of a public API and therefore users cannot benefit from it

src/DependencyInjection/Compiler/DbalTracingPass.php
```diff
@@ -92,4 +96,24 @@ private function getSetMiddlewaresMethodCallArguments(Definition $definition): a
...
+     */
+    private function assertRequiredDbalVersion(): void
+    {
+        if (interface_exists(Driver::class)) {
```
ste93cry, line 105: I think we are safe in removing this condition because the nested statements works on their own

ste93cry (changes requested):
Two small nitpicks and then LGTM for me

ste93cry (approved):

PR #541, opened 2021-08-16
fix(types): added nullability to getDatabase method

santoslluis:
If getDatabase doesn't allow null return values, it will fail when the Driver returns null for this method.

It happens in https://github.com/webmozart/doctrine-dbal/blob/master/lib/Doctrine/DBAL/Driver/PDOSqlite/Driver.php#L114 when using `DATABASE_URL="sqlite:///:memory:"`

this is weird because in-memory dbs are usually used in a test environments were sentry is deactivated, but this class is still being loaded and throws errors in all the tests. 

I get the following error in all the tests:
`TypeError : Return value of Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriver::getDatabase() must be of the type string, null returned`

with this change it works for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix return type for `TracingDriver::getDatabase()` method (#541)
+
 ## 4.2.0 (2021-08-12)
 
 - Log the bus name, receiver name and message class name as event tags when using Symfony Messenger (#492)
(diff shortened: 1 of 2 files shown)


ste93cry:
Thank you for catching this bug, I just take a look at the PHPDoc of that method and it doesn't mention `null` so that's why it was like it was. Can you please add an entry to the CHANGELOG with a reference to this PR 🙏?

santoslluis:
> Thank you for catching this bug, I just take a look at the PHPDoc of that method and it doesn't mention `null` so that's why it was like it was. Can you please add an entry to the CHANGELOG with a reference to this PR pray?

sure

ste93cry (approved):
Thank you for the contribution, LGTM 🥳

PR #548, opened 2021-08-20
Fix missing instrumentation of the `Statement::execute()` method of Doctrine DBAL

### LINKED ISSUE #544: Prepared queries when using Doctrine DBAL are not instrumented
I'm using bundle 4.2.0 and using tracing for sentry.

When some SQL is used for selecting some data in sentry I see only a trace for sql.conn.prepare. But this does not show the time spent for selecting data itself.

Maybe there is a way to include such information in trace too? I want to see in trace how much time was spent for getting data from db

Thank you.

ste93cry:
Fixes #544: thanks to @jokaorgua I found out that the `Doctrine\DBAL\Driver\Statement::execute()` method has to be instrumented as well for the tracing of SQL queries to work properly in case a prepared statement is used, e.g. when calling the `Doctrine\DBAL\Connection::fetch*()` methods and passing some query params to them. I voluntarily avoided a child span for the `TracingStatement::execute()` method because you may prepare a query and run it later, so it doesn't make sense to have it nested into the one of the `TracingDriverConnection::prepare()` method as it would have happened by creating it inside the callback of the `TracingDriverConnection::traceFunction()` function


Jean85 (approved):
This seems a bit complicated, but we don't have any other way around it...

Issue #545, opened 2021-08-19
Dependency conflict with `symfony/cache-contracts` when updating from `4.1.2` to `4.2` & using non flex symfony 4.4 project


nocive:
Perhaps I'm missing something obvious here, so apologies upfront if that's the case.
Happy to provide more information if necessary.

### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
PHP 7.4.x
sentry-php 3.3.2
sentry-symfony 4.1.4 & 4.2

### Steps to Reproduce

1. Using 4.1.4
```
$ composer show sentry/*
sentry/sdk            3.1.0 This is a metapackage shipping sentry/sentry with a recommended HTTP client.
sentry/sentry         3.3.2 A PHP SDK for Sentry (http://sentry.io)
sentry/sentry-symfony 4.1.2 Symfony integration for Sentry (http://getsentry.com)
```

2. Attempt to update to 4.2 (using `-W` makes no difference)
```
$ composer update sentry/sentry-symfony:^4.2
Loading composer repositories with package information
Updating dependencies                                 
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - sentry/sentry-symfony[dev-master, 4.2.0] require symfony/cache-contracts ^2.4 -> found symfony/cache-contracts[dev-main, v2.4.0, 2.4.x-dev, 2.5.x-dev (alias of dev-main)] but these were not loaded, likely because it conflicts with another require.
    - sentry/sentry-symfony 4.2.x-dev is an alias of sentry/sentry-symfony dev-master and thus requires it to be installed too.
    - Root composer.json requires sentry/sentry-symfony ^4.1, ^4.2 -> satisfiable by sentry/sentry-symfony[4.2.0, 4.2.x-dev (alias of dev-master)].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

Symfony dependencies:
```
$ composer show symfony/*
symfony/contracts                v1.1.10 A set of abstractions extracted out of the Symfony components
symfony/deprecation-contracts    v2.4.0  A generic function and convention to trigger deprecation notices
symfony/monolog-bundle           v3.7.0  Symfony MonologBundle
symfony/phpunit-bridge           v5.3.4  Provides utilities for PHPUnit, especially user deprecation notices management
symfony/polyfill-ctype           v1.23.0 Symfony polyfill for ctype functions
[...]

PR #549, opened 2021-08-24
Lower the version requirement for the `symfony/cache-contracts` package

ste93cry:
While investigating #545, I noticed that we are requiring `symfony/cache-contracts:^2.4` while `symfony/symfony:^4.4` requires `symfony/cache-contracts:^1.1` (actually it replaces that package with `symfony/contracts:^1.1`). This makes it impossible to install `sentry/sentry-symfony:^4.2` on a plain Symfony `4.4` installation (no Flex). ~However, I cannot understand why it works with `sentry/sentry-symfony:4.1.3` though, since that version is the one where the `composer.json` requirement actually got changed~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "jean85/pretty-package-versions": "^1.5 || ^2.0",
         "php-http/discovery": "^1.11",
         "sentry/sdk": "^3.1",
-        "symfony/cache-contracts": "^2.4",
+        "symfony/cache-contracts": "^1.1||^2.4",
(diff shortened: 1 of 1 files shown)


nocive (approved):

Jean85 (approved):
[Changelog](https://github.com/symfony/contracts/blob/main/CHANGELOG.md#200) doesn't list anything problematic, LGTM :+1:

nocive:
thanks guys :+1:

PR #552, opened 2021-09-11
define TracingStatement alias conditionally

yivi:
Since the alias is defined unconditionally, when using opcache PHP emits a Warning that can break applications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


ste93cry:
Can you please add a CHANGELOG entry linking to this PR and describing the reason for the fix?

yivi:
Added a CHANGELOG entry on the "Unreleased" section.

ste93cry (approved):

Issue #484, opened 2021-04-19
New doctrine/dbal polyfills have side effects


dFayet:
Since sentry/sentry-symfony version 4.1.0,  new doctrine/dbal polyfills were added:

https://github.com/getsentry/sentry-symfony/blob/b2500fb71a7acef734ba72f8d0f98ee530cf8174/src/aliases.php#L97-L120
Related PR: #423 

This has the effect of misleading code based on the existence, or not, of these doctrine/dbal v2 interfaces.
As basing the detection only on that is maybe not a great idea, creating polyfill for 3rd party library seems to be quite odd too.

Jean85:
I agree that this approach is not great; we did something similar with the Symfony classes, in the first part of the file, but in that case we declared completely new names to avoid clashing.

@ste93cry do you think that we can do the same here?

ste93cry:
@Jean85 this looks (at least partially) solved by #527, right? As far as I see only two aliases remain

Jean85:
It should. Both aliases are against a classname that it's our own, so we shouldn't have this issue anymore.

@dFayet can you test our `develop` branch to confirm?

dFayet:
@Jean85 it reduces the numbers of errors possible, but it does not solve the issue as the alias `class_alias(ExceptionConverterDriverInterface::class, LegacyExceptionConverterDriverInterface::class); ` is still here.

ste93cry:
>it does not solve the issue as the alias

Yep, that was my thinking. But to be honest I don't see how we could get rid of the remaining ones, we have to implement the interface and in this case it's Symfony not making his own business

dFayet:
> it's Symfony not making his own business
I don't think this is Symfony's business here, we have problems with Doctrine.

Is the alias needed somewhere else than in `Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriver` ?
If no, wouldn't it be possible to have 2 implementations like `Dbal2TracingDriver` and `Dbal3TracingDriver` (with abstraction, trait, ...)?

Jean85:
@dFayet you're welcome to try it if you want, PR are welcome.
You should target `develop` though, since #527 changed so much already there.

PR #553, opened 2021-09-11
Improve the compatibility layer for Doctrine DBAL to avoid side-effects of polyfilling

ste93cry:
This is an attempt at fixing #483 and #484. I decided to split the `TracingDriver` class into two versions, each compatible with just one of the major versions of Doctrine DBAL we support (`2.x` and `3.x`)


ste93cry:
After thinking again about this, and while we have some clear advantages of splitting the implementation into two, e.g. the fact that we can add typehints to more things and evolve the two classes independently without worrying about keeping the compatibility with both major versions at the same time, I'm pretty sure this does not solve #483 and there is no way to solve it. Basically, the [`DebugClassLoader`](https://github.com/symfony/error-handler/blob/5.3/DebugClassLoader.php) of Symfony blindly parse the annotations on the classes that it loads and as soon as it finds one that is marked as deprecated and that is loaded from something outside of its vendor it reports the deprecation. In our case, we need to implement those deprecated interfaces to expose the fact that the decorated driver implements them, since it's quite common to use an `instanceof` check in this context. I still think that my comment about Symfony not making his own business applies perfectly in this case, the thing that the `DebugClassLoader` is doing would be better suited for a static analysis tool which has far more understanding of the context around how the code is called and how it works

Jean85 (changes requested):
LGTM :+1: splitting is surely a better approach.

Maybe we should declare those implementations `@internal`?

ste93cry:
>Maybe we should declare those implementations `@internal`?

Yes, that's a good idea!

Jean85 (approved):

starred-gijs:
This PR fixed it already I guess https://github.com/getsentry/sentry-symfony/pull/552

Jean85:
#552 is released as 4.2.3, you can test it out.

PR #555, opened 2021-09-13
Pass logger from options to TransportFactory

simPod:
I have noticed that a transport error was being logged in https://github.com/getsentry/sentry-php/blob/bfab90c7ef921221b52b75b5108201e87f0f4661/src/Transport/HttpTransport.php#L112-L115

but there always was default NullLogger registered. So I've found that even though I set logger via `options` it was not being passed to the Transport.

This fixes it.

### DIFF
--- a/src/DependencyInjection/SentryExtension.php
+++ b/src/DependencyInjection/SentryExtension.php
@@ -27,6 +27,7 @@ use Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriverMiddleware;
 use Sentry\SentryBundle\Tracing\Twig\TwigTracingExtension;
 use Sentry\Serializer\RepresentationSerializer;
 use Sentry\Serializer\Serializer;
+use Sentry\Transport\TransportFactoryInterface;
 use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Cache\CacheItem;
 use Symfony\Component\Config\FileLocator;
@@ -124,6 +125,13 @@ final class SentryExtension extends ConfigurableExtension
             ->setPublic(false)
             ->setArgument(0, new Reference('sentry.client.options'));
 
+        $loggerReference = null === $config['logger']
+            ? new Reference(NullLogger::class, ContainerBuilder::IGNORE_ON_INVALID_REFERENCE)
+            : new Reference($config['logger']);
+
+        $factoryBuilderDefinition = $container->getDefinition(TransportFactoryInterface::class);
+        $factoryBuilderDefinition->setArgument('$logger', $loggerReference);
+
         $clientBuilderDefinition = (new Definition(ClientBuilder::class))
             ->setArgument(0, new Reference('sentry.client.options'))
             ->addMethodCall('setSdkIdentifier', [SentryBundle::SDK_IDENTIFIER])
@@ -131,7 +139,7 @@ final class SentryExtension extends ConfigurableExtension
             ->addMethodCall('setTransportFactory', [new Reference($config['transport_factory'])])
             ->addMethodCall('setSerializer', [$serializer])
             ->addMethodCall('setRepresentationSerializer', [$representationSerializerDefinition])
-            ->addMethodCall('setLogger', [null !== $config['logger'] ? new Reference($config['logger']) : new Reference(NullLogger::class, ContainerBuilder::IGNORE_ON_INVALID_REFERENCE)]);
+            ->addMethodCall('setLogger', [$loggerReference]);
 
         $container
             ->setDefinition('sentry.client', new Definition(Client::class))
--- a/tests/DependencyInjection/SentryExtensionTest.php
+++ b/tests/DependencyInjection/SentryExtensionTest.php
@@ -26,6 +26,7 @@ use Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriverMiddleware;
 use Sentry\SentryBundle\Tracing\Twig\TwigTracingExtension;
 use Sentry\Serializer\RepresentationSerializer;
 use Sentry\Serializer\Serializer;
+use Sentry\Transport\TransportFactoryInterface;
 use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Console\ConsoleEvents;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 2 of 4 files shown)


ste93cry (changes requested):

tests/DependencyInjection/SentryExtensionTest.php
```diff
@@ -245,6 +246,15 @@ public function testClientIsCreatedFromOptions(): void
         $this->assertEquals($methodCalls[4][1][0]->getArgument(0), new Reference('sentry.client.options'));
     }
 
+    public function testLoggerIsPassedToTransportFactory(): void
+    {
+        $container = $this->createContainerFromFixture('full');
+
+        $transportFactoryDefinition = $container->findDefinition(TransportFactoryInterface::class);
+        $logger = $transportFactoryDefinition->getArgument('$logger');
+        $this->assertSame('app.logger', $logger->__toString());
```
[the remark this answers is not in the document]
simPod: Do you want to magically cast it? It's hard for SA to understand and detects bug.

This guy is right actually https://github.com/ShittySoft/symfony-live-berlin-2018-doctrine-tutorial/pull/3#issuecomment-460441229

src/DependencyInjection/SentryExtension.php
```diff
@@ -124,19 +130,35 @@ private function registerConfiguration(ContainerBuilder $container, array $confi
...
             ->setArgument(0, new Reference('sentry.client.options'))
             ->addMethodCall('setSdkIdentifier', [SentryBundle::SDK_IDENTIFIER])
             ->addMethodCall('setSdkVersion', [PrettyVersions::getVersion('sentry/sentry-symfony')->getPrettyVersion()])
             ->addMethodCall('setTransportFactory', [new Reference($config['transport_factory'])])
             ->addMethodCall('setSerializer', [$serializer])
             ->addMethodCall('setRepresentationSerializer', [$representationSerializerDefinition])
-            ->addMethodCall('setLogger', [null !== $config['logger'] ? new Reference($config['logger']) : new Reference(NullLogger::class, ContainerBuilder::IGNORE_ON_INVALID_REFERENCE)]);
+            ->addMethodCall('setLogger', [$loggerReference]);
+
+        $factoryBuilderDefinition = (new Definition(TransportFactory::class))
```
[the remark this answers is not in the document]
simPod: I'll try tomorrow but not sure if it's possible as I'm not very familiar with SF DI configuration. `$config['logger']` is optional so not sure if I'll be able to pass it.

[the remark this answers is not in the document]
ste93cry: I think that something along the lines of this should work:

```php
$factoryBuilderDefinition  = $this->getDefinition(TransportFactoryInterface::class);
$factoryBuilderDefinition->setArgument('$logger', ...);
```

tests/DependencyInjection/SentryExtensionTest.php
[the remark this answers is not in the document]
ste93cry: Makes sense, I'm used to the cast but no strong opinion on leaving this as is, it was mainly for a matter of consistency with other points in the code where we do the same and because I don't really like to call a magic method manually

src/DependencyInjection/SentryExtension.php
```diff
@@ -36,6 +37,10 @@
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\ErrorHandler\Error\FatalError;
 use Symfony\Component\HttpKernel\DependencyInjection\ConfigurableExtension;
+use function array_filter;
+use function array_map;
+use function array_unshift;
+use function class_exists;
```
ste93cry, line 43: I don't know why PHPCS didn't report this, maybe it's configured to not warn if importing functions, but usually we don't do it. :pray: revert these changes back

src/DependencyInjection/SentryExtension.php
```diff
@@ -124,14 +129,22 @@ private function registerConfiguration(ContainerBuilder $container, array $confi
             ->setPublic(false)
             ->setArgument(0, new Reference('sentry.client.options'));
 
+        $loggerReference = null === $config['logger']
+            ? new Reference(NullLogger::class, ContainerBuilder::IGNORE_ON_INVALID_REFERENCE)
+            : new Reference($config['logger']);
+
+        $factoryBuilderDefinition = $container->getDefinition(TransportFactoryInterface::class);
+        $factoryBuilderDefinition->setArgument('$httpClient', null);
```
ste93cry, line 137: What's the reason for also setting this argument explicitly?

tests/DependencyInjection/SentryExtensionTest.php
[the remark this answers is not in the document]
ste93cry: Can you please still add the blank line before the assertion 🙏 to make the code a little more readable at first sight?

ste93cry (changes requested):
Would you mind updating the CHANGELOG too to reference this PR?

src/DependencyInjection/SentryExtension.php
simPod: Symfony tells me to do so: Cannot set sixth without setting fifth

ste93cry: Makes sense, but then I think that as it's done in some parts of Symfony itself it would be better to declare all arguments in the XML config and then override here the last one. WDYT?

src/DependencyInjection/SentryExtension.php
simPod: there's cs fixer, not phpcs. I use phpcs where it's configurable, not sure about fixer 🤷🏾‍♀️

src/DependencyInjection/SentryExtension.php
simPod: smart "hack"

simPod:
ah cahngelog

src/DependencyInjection/SentryExtension.php
ste93cry: By setting the `$httpClient` argument to a service, the behaviour of the existing applications may change if that service exists because it may be different from the one that was returned while autodiscovering the HTTP client. To be safe, this should work as expected:

```xml
<service id="Sentry\Transport\TransportFactoryInterface" class="Sentry\SentryBundle\Transport\TransportFactory">
    ...
    <argument>null</argument>
    <argument /> <!-- $logger -->
</service>
```

simPod: aight, i'm no symfony expert so I went with what you've suggested

ste93cry (approved):
Everything looks good, thank you for the contribution 💪

PR #562, opened 2021-10-01
Add failing test for register_error_listener

jrjohnson:
Pulling this value from the ENV doesn't seem to work as this test
demonstrates.

Instead of a boolean I'm seeing `env_e71e97ef5efbff51_bool_REGISTER_ERROR_LISTENER_ee2aa964b5d330a7af36cf7065e7419f` at https://github.com/getsentry/sentry-symfony/blob/31edf15fd22fcfb2ba0989cef7d70f34a41f6abf/src/DependencyInjection/SentryExtension.php#L259 I assume something needs to be adjusted to get Symfony to send the real value, but I'm not familiar with the DI system and can't see where that may be so I've just included this failing test in the hope that it helps resolve the issue.

### DIFF
--- a/tests/DependencyInjection/Fixtures/php/error_listener_disabled_from_env.php
+++ b/tests/DependencyInjection/Fixtures/php/error_listener_disabled_from_env.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/** @var ContainerBuilder $container */
+$container->setParameter('env(REGISTER_ERROR_LISTENER)', 'false');
+$container->loadFromExtension('sentry', [
+    'register_error_listener' => '%env(bool:REGISTER_ERROR_LISTENER)%',
+]);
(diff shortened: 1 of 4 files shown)


Jean85:
Thank you for reporting the issue and giving us a regression test! Very much appreciated!

@ste93cry I didn't know where to start about this, but I've seen from Doctrine Bundle that they use the Symfony Config Processor: https://github.com/doctrine/DoctrineBundle/blob/62a188ce2192e6b3b7a2019c26c5001778818b83/DependencyInjection/DoctrineExtension.php#L74-L75

So we would need a rehaul of our `SentryExtension` class... WDYT? Should we do it?

ste93cry:
It looks like you're trying to configure whether the error listener should be registered at runtime rather at compile time. However, and this is something not specifically tied to this bundle but that don't work also for almost all the other bundles out of here, the configuration must be parsed at compile time because the container is dumped to a file. Since this config option actually modifies which services get registered, what would happen is that you would know too late if the service should be registered and or not in the service container

jrjohnson:
Ah, that makes sense @ste93cry. I do want to be able to activate this at runtime for only some of our partner setups. I'll go back to my original plan to use the DSN config value to do this. Thanks. I'll go ahead and close this PR because it sounds like this issue is more of a design decision and requirement than a bug.

Issue #488, opened 2021-04-22
Sentry bundle breaks doctrine migrations.


vuryss:
### Environment

Latest symfony 5.2 + doctrine migrations 3.1 + sentry-symfony 4.1

### Steps to Reproduce

Having the above combination completely breaks the Doctrine migrations unless we add to each migrations a method indicating that the migration is not transactional.
Doctrine had this issue before but they fixed it with this check:

https://github.com/doctrine/migrations/blob/96c64fa17477e503da8245f14541d70d7ae76091/lib/Doctrine/Migrations/Tools/TransactionHelper.php#L39

However, sentry decorates the connection with:

`Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriverConnection`  

which is obviously not a PDO instance, so the aforementioned check fails and we get a nasty error:

```
In TracingDriverConnection.php line 163:
                                  
  [PDOException]                  
  There is no active transaction  
```                                  

### Expected Result

Well obviously I expect migrations to work with and without the sentry bundle :)

### Actual Result

```
In TracingDriverConnection.php line 163:
                                  
  [PDOException]                  
  There is no active transaction  
```

ste93cry:
I think there is nothing we can do here for the moment, since the check done in the Doctrine repo is clearly intended to work only for PDO-based drivers. @greg0ire is there any reason for that check to be like that? Maybe we could make the `inTransaction` method part of the `Doctrine\DBAL\Driver\Connection` interface so that the check in `TransactionHelper` does not have to care whether or not the connection is a PDO connection, but it just has to ask whether the transaction is active or not. Each driver connection can then report back either `true` or `false` based on specific conditions. Otherwise, any attempt to wrap the driver's connection will cause the same issue we're having here

greg0ire:
I think the reason was we needed a fix that would not require bumping our dependency on the DBAL, but you are right about `inTransaction()` possibly being the right API: https://github.com/doctrine/dbal/pull/4544#issuecomment-799596645

ste93cry:
Understood. Do you think it would be feasible to implement this API in both version 2.x and 3.x? And do you think there would be anyone in the Doctrine team interested in doing this?

greg0ire:
> Do you think it would be feasible to implement this API in both version 2.x and 3.x?

Feasible, sure, but I think you and I both know it will not be accepted: https://github.com/doctrine/dbal/issues/4495

[...]

PR #567, opened 2021-10-16
Fix decoration of the Doctrine DBAL connection when it implemented the `ServerInfoAwareConnection` interface

ste93cry:
This solves https://github.com/getsentry/sentry-symfony/issues/488#issuecomment-941401855: while instrumenting the Doctrine DBAL connection we weren't mirroring whether the original connection had implemented the `ServerInfoAwareConnection` interface, causing issues with `instanceof` checks


ste93cry:
@fritzmg would you be so gentle  to try this out and see if it works?

PR #573, opened 2021-11-19
Add the Monolog hook functionality

Jean85:
This is another attempt at #337 

I'm currently following the strategy outlined in https://github.com/getsentry/sentry-symfony/issues/337#issuecomment-800920045

It seems to work with this config:
```yaml
sentry:
  register_error_listener: false
  monolog:
    error_handler:
      enabled: true
  options:
    # TODO: add a simple option to switch off the error/fatal integrations
    default_integrations: false
    integrations: []
```
But E2E tests are failing and I'm having a hard time to run them with the debugger.

### DIFF
--- a/src/DependencyInjection/Compiler/RegisterMonologHandlerPass.php
+++ b/src/DependencyInjection/Compiler/RegisterMonologHandlerPass.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\SentryBundle\DependencyInjection\Compiler;
+
+use Sentry\Monolog\Handler;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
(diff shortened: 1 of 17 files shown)


Jean85:
Reason found: attacching the handler to `monolog.logger` is not enough, that's just the logger for the default channel (normally `app`).

We would need to add the handler to all the existing channels, but I can't find an easy way to do it, the services are not tagged, they are just called `monolog.logger.{channel_name}`.

cbastienbaron:
hey @Jean85 , 

thanks for this pr / explanation :) from what i understand, at this time if we want to work with monolog and sentry (as simple handler of monolog) , the best solution (for prod env) is to use alt https://github.com/B-Galati/monolog-sentry-handler/ ? (the time to this pr to be finished :crossed_fingers: )

Issue #576, opened 2021-11-30
Declaration of TraceableCacheAdapter::getItem($key) must be compatible with Symfony\Component\Cache\Adapter\AdapterInterface::getItem(mixed $key)


jaras2678:
Symfony ver. 5.4.0
Sentry-symfony ver. 4.2.4

```
Symfony\Component\ErrorHandler\Error\FatalError^ {#8863
  -error: array:4 [
    "type" => 64
    "message" => "Declaration of Sentry\SentryBundle\Tracing\Cache\TraceableCacheAdapter::getItem($key) must be compatible with Symfony\Component\Cache\Adapter\AdapterInterface::getItem(mixed $key): Symfony\Component\Cache\CacheItem"
    "file" => "./vendor/sentry/sentry-symfony/src/Tracing/Cache/TraceableCacheAdapter.php"
    "line" => 38
  ]
  #message: "Compile Error: Declaration of Sentry\SentryBundle\Tracing\Cache\TraceableCacheAdapter::getItem($key) must be compatible with Symfony\Component\Cache\Adapter\AdapterInterface::getItem(mixed $key): Symfony\Component\Cache\CacheItem"
  #code: 0
  #file: "./vendor/sentry/sentry-symfony/src/Tracing/Cache/TraceableCacheAdapter.php"
  #line: 38
```

PR #577, opened 2021-11-30
TraceableCacheAdapter compatible with Symfony\Component\Cache\Adapter…

jaras2678:
Issuse https://github.com/getsentry/sentry-symfony/issues/576

### DIFF
--- a/src/Tracing/Cache/TraceableCacheAdapterTrait.php
+++ b/src/Tracing/Cache/TraceableCacheAdapterTrait.php
@@ -35,7 +35,7 @@ trait TraceableCacheAdapterTrait
     /**
      * {@inheritdoc}
      */
-    public function getItem($key)
+    public function getItem(mixed $key)
     {
         return $this->traceFunction('cache.get_item', function () use ($key) {
             return $this->decoratedAdapter->getItem($key);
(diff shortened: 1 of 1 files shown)


ste93cry:
This change is a breaking change. Moreover, it doesn't work on PHP 7 because the `mixed` typehint was introduced in version 8. There's already an open PR to support Symfony 6, I will find a way to integrate this change there

src/Tracing/Cache/TraceableCacheAdapterTrait.php
```diff
@@ -35,7 +36,7 @@ trait TraceableCacheAdapterTrait
     /**
      * {@inheritdoc}
      */
-    public function getItem($key)
+    public function getItem(mixed $key): CacheItem
```
Jean85, line 39: The variance rules are reversed for arguments, so just by doing this we should be fine:
```suggestion
    public function getItem($key): CacheItem
```

Jean85 (changes requested):
Also, please add a changelog entry. I would also require a test, but I would expect this to land from #566. @ste93cry can you check on this?

ste93cry:
@Jean85 there's no need to do anything here, as I said before there's already an open PR to support Symfony 6, with this change included

PR #578, opened 2021-12-02
Update PHPStan and PHP-CS-Fixer to their latest version

ste93cry:
As per title, I've updated PHPStan and PHP-CS-Fixer to their latest version. A good bunch of errors that are currently ignored in the PHPStan baseline can be removed once the new version with the support for the `key-of` type gets released.


Jean85 (approved):
LGTM :+1:

.github/workflows/static-analysis.yaml
```diff
@@ -18,8 +18,6 @@ jobs:
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
-        with:
-          php-version: '8.0'
```
Jean85, line 22: As in https://github.com/getsentry/sentry-php/pull/1272, I would prefer to have a pinned version to avoid issues every november.

.github/workflows/static-analysis.yaml
```diff
@@ -52,8 +50,6 @@ jobs:
...
```
Jean85, line 56: Here too

Jean85 (changes requested):

Jean85 (approved):

PR #588, opened 2022-01-03
Support version 3 of the `symfony/cache-contracts` package

ste93cry:
This PR adds support for `symfony/cache-contracts` package version `3.x`. Since one of the interfaces is using the `mixed` keyword which has been introduced in PHP 8, we have to fork the implementation of our classes to support both the new and the old versions

### DIFF
--- a/src/Tracing/Cache/TraceableCacheAdapterForV2.php
+++ b/src/Tracing/Cache/TraceableCacheAdapterForV2.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\SentryBundle\Tracing\Cache;
+
+use Sentry\State\HubInterface;
+use Symfony\Component\Cache\Adapter\AdapterInterface;
+use Symfony\Component\Cache\PruneableInterface;
+use Symfony\Component\Cache\ResettableInterface;
+use Symfony\Contracts\Cache\CacheInterface;
+
+/**
+ * This implementation of a cache adapter supports the distributed tracing
+ * feature of Sentry.
+ *
+ * @internal
+ */
+final class TraceableCacheAdapterForV2 implements AdapterInterface, CacheInterface, PruneableInterface, ResettableInterface
+{
+    /**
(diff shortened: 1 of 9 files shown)


tests/DependencyInjection/Compiler/CacheTracingPassTest.php
```diff
@@ -50,7 +50,7 @@ public function processDataProvider(): \Generator
                     ->setPublic(true)
                     ->addTag('cache.pool'),
             ],
-            TraceableCacheAdapter::class,
+            TraceableCacheAdapterForV2::class,
```
Jean85, line 53: Why do we never get V3 here?

ste93cry: Ops, when I renamed the class using the IDE this got changed and I didn't pay attention to it. Fixed

src/Tracing/Cache/TraceableCacheAdapterForV2.php
```diff
@@ -0,0 +1,53 @@
...
+     * {@inheritdoc}
+     *
+     * @param mixed[] $metadata
+     *
+     * @return mixed
+     */
+    public function get(string $key, callable $callback, float $beta = null, array &$metadata = null)
+    {
+        return $this->traceFunction('cache.get_item', function () use ($key, $callback, $beta, &$metadata) {
+            if (!$this->decoratedAdapter instanceof CacheInterface) {
```
Jean85, line 46: Here we're saying that we're not sure to have the `CacheInterface`, but we're declaring it; isn't this a risk for stuff like autoloading?

ste93cry: I'm not sure I get what you mean. The adapter does not necessarily implements the `CacheInterface` method, therefore we have to check that it does before calling `get()` on it. What does autoloading have to do with this?

Jean85: Sorry, I meant autowire: our wrapper implements always the interface, and that could confuse the autowire since the decorated service could not.

Jean85 (approved):

Issue #592, opened 2022-01-19
How do I set the connect timeout and timeout values?


matthew-gill:
We're in the process of implementing Sentry for our Symfony app and would like to set some v. conservative timeouts in the event that the Sentry server is unavailable or slow to respond.

I've been able to set

```yml
sentry:
  options:
    send_attempts: 1
```

Which makes a slight bit of difference, but it seems as if the `DEFAULT_HTTP_CONNECT_TIMEOUT` and `DEFAULT_HTTP_TIMEOUT` are hard coded in `vendor/sentry/sentry/src/HttpClient/HttpClientFactory.php
vendor/sentry/sentry/src/HttpClient/HttpClientFactory.php`

Is there a way I'm missing to override this? Perhaps using the `transport` option (but I can't seem to find any documentation for it).

If you have an example of how to use the `transport` option that would be great. Otherwise, I'll try to raise a PR to set these two values by the `option` config.

matthew-gill:
I've now raised both PRs above. I'm not sure if this is the way to do it, if you have a better way (or it's already possible) I'd love to know, thanks!

Jean85:
I thought that this feature was not feasible since it would require handling that setting for each proper client implementation but, since we're already doing it (but with a hard coded value) I think it has its merits. @ste93cry do you agree?

I have some additional thoughts:
 * it's a new feature, so both PRs have to target develop
 * here on Symfony we will need to lock onto that newer `sentry/sentry` package version, or it will break
 * there are some other corrections needed on the PRs, but I'll leave a contextual review there.

Thanks for the contribution!

ste93cry:
@ste93cry do you agree?

Yes, let's do it once for all

Jean85:
Now both PRs are released! Thank you @matthew-gill!

PR #593, opened 2022-01-19
Support http_connect_timeout and http_timeout options

matthew-gill:
I raised this issue: https://github.com/getsentry/sentry-symfony/issues/592

This PR adds the ability to set options for http (timeout) options, when paired with https://github.com/getsentry/sentry-php/pull/1282 (which modifies the `sentry-php` SDK) will allow the setting of these options via Symfony yml configuration.

Example use:

`sentry.yml`:
```yml
sentry:
    options:
        http_timeout: 1
        http_connect_timeout: 1
```

If I've missed the mark, feel free to reject - I'd just love to be able to do this in my app, if it's possible already please let me know, thanks!

### DIFF
--- a/tests/DependencyInjection/Fixtures/php/full.php
+++ b/tests/DependencyInjection/Fixtures/php/full.php
@@ -36,11 +36,11 @@ $container->loadFromExtension('sentry', [
         'send_default_pii' => true,
         'max_value_length' => 255,
         'http_proxy' => 'proxy.example.com:8080',
+        'http_timeout' => 10,
+        'http_connect_timeout' => 15,
         'capture_silenced_errors' => true,
         'max_request_body_size' => 'none',
         'class_serializers' => ['App\\FooClass' => 'App\\Sentry\\Serializer\\FooClassSerializer'],
-        'http_timeout' => 10,
-        'http_connect_timeout' => 15,
     ],
     'messenger' => [
         'enabled' => true,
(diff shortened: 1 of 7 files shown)


Jean85 (changes requested):
This needs:
 * to lock against `sentry/sentry` which includes https://github.com/getsentry/sentry-php/pull/1282 (probably `^3.4`?)
 * a changelog entry
 * additions to the tests that cover the configuration

tests/DependencyInjection/Fixtures/php/full.php
```diff
@@ -39,6 +39,8 @@
         'capture_silenced_errors' => true,
         'max_request_body_size' => 'none',
         'class_serializers' => ['App\\FooClass' => 'App\\Sentry\\Serializer\\FooClassSerializer'],
+        'http_timeout' => 10,
+        'http_connect_timeout' => 15,
```
Jean85, line 43: As in https://github.com/getsentry/sentry-php/pull/1282#discussion_r787947940, you can't set a default here or you'll trigger errors if there is no default client available.

Jean85 (changes requested):

tests/DependencyInjection/Fixtures/php/full.php
matthew-gill: If I set them as null I get 

```
Symfony\Component\Config\Definition\Exception\InvalidTypeException : Invalid type for path "sentry.options.http_timeout". Expected "int", but got "null".
```

How do you suggest to fix that? Would it be removing it from the fixture file and just creating a fixture which contains those options?

matthew-gill: For what it's worth, the fixture file does seem to contain a default value for `http_proxy`:

```
        'http_proxy' => 'proxy.example.com:8080',
```

Jean85:
Build restarted, we have Psalm and a couple of other jobs failing.

ste93cry:
I rebased the PR on the latest `develop` and the situation is slightly better, but I can't understand why two jobs are still failing. Do you have any idea?

ste93cry (approved):

Jean85 (approved):

PR #613, opened 2022-04-05
Default Statement::bindParam $length to 0, as using null causes a deprecation with PHP8+

### LINKED ISSUE #611: Calling PDOStatement::bindParam with only 2 params should not trigger a deprecation
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
sentry/sentry-symfony 4.2.7 with Doctrine DBAL 2.13.8

### Steps to Reproduce

Call PDOStatement bindParam with only 2 params:
```php
$statement = $this->connection->prepare('...');
$statement->bindParam(':param1', $values['param1']);
```

### Expected Result

No exception triggered

### Actual Result

ErrorException: Deprecated: PDOStatement::bindParam(): Passing null to parameter #4 ($maxLength) of type int is deprecated

PR https://github.com/getsentry/sentry-symfony/pull/586 "fixed" this issue already, but it's still broken when using DBALv2.

In particular, if only 2 arguments are passed to the method, then the `\array_slice(\func_get_args(), 3)` code will return an empty array, which results in `null` being used for `$length` as it's the default. In DBALv2, https://github.com/doctrine/dbal/blob/2.13.8/lib/Doctrine/DBAL/Driver/PDO/Statement.php simply extends PDOStatement, with no additional functionality. So there's nothing to catch the null `$length` and convert it to 0 (the default in PDO), as was done in DBALv3.

The solution would be implementing a similar code change as DBALv3 does:
```php
    public function bindParam($param, &$variable, $type = ParameterType::STRING, $length = null): bool
    {
        return $this->decoratedStatement->bindParam($param, $variable, $type, $length ?? 0, ...\array_slice(\func_get_args(), 4));
    }
```

cjunge-work:
When using PHP8+ with Doctrine DBALv2 and calling `PDOStatement::bindParam()` with just 2 params causes a deprecation for passing null to an int param. The default in PDO is `0`, so this change defaults to 0 when calling the decorated statement.

Fixes #611

### DIFF
--- a/tests/Tracing/Doctrine/DBAL/TracingStatementForV2Test.php
+++ b/tests/Tracing/Doctrine/DBAL/TracingStatementForV2Test.php
@@ -215,6 +215,10 @@ final class TracingStatementForV2Test extends DoctrineTestCase
 if (!interface_exists(Statement::class)) {
     abstract class TracingStatementForV2Stub
     {
+        /**
+         * @var int|null
+         */
+        public $bindParamLengthValue;
     }
 } else {
     /**
@@ -226,7 +230,10 @@ if (!interface_exists(Statement::class)) {
          * @var int
          */
         public $bindParamCallArgsCount = 0;
-        public $bindParamLengthValue = null;
+        /**
+         * @var int|null
+         */
+        public $bindParamLengthValue;
 
         public function bindParam($param, &$variable, $type = ParameterType::STRING, $length = null): bool
         {
--- a/tests/Tracing/Doctrine/DBAL/TracingStatementForV2Test.php
+++ b/tests/Tracing/Doctrine/DBAL/TracingStatementForV2Test.php
@@ -123,7 +123,7 @@ final class TracingStatementForV2Test extends DoctrineTestCase
      * @param mixed[] $callArgs
      * @param mixed[] $expectedCallArgs
      */
-    public function testBindParam(array $callArgs, int $expectedLengthArg, array $expectedCallArgs): void
+    public function testBindParam(array $callArgs, array $expectedCallArgs): void
     {
         $variable = 'bar';
         $decoratedStatement = $this->createPartialMock(TracingStatementForV2Stub::class, array_diff(get_class_methods(TracingStatementForV2Stub::class), ['bindParam']));
@@ -132,7 +132,6 @@ final class TracingStatementForV2Test extends DoctrineTestCase
 
         $this->assertTrue($this->statement->bindParam('foo', $variable, ParameterType::INTEGER, ...$callArgs));
         $this->assertSame($expectedCallArgs, $decoratedStatement->bindParamCallArgs);
-        $this->assertSame($expectedLengthArg, $decoratedStatement->bindParamLengthValue);
     }
 
     /**
@@ -142,19 +141,16 @@ final class TracingStatementForV2Test extends DoctrineTestCase
     {
         yield '$length parameter not passed at all' => [
             [],
-            0,
             ['foo', 'bar', 1, 0],
         ];
 
         yield '$length parameter passed as `null`' => [
             [null],
-            0,
             ['foo', 'bar', 1, 0],
         ];
 
         yield 'additional parameters passed' => [
             [null, 'baz'],
-            0,
             ['foo', 'bar', 1, 0, 'baz'],
         ];
     }
@@ -234,11 +230,6 @@ if (!interface_exists(Statement::class)) {
          * @var mixed[]
          */
         public $bindParamCallArgs = [];
-
-        /**
-         * @var int|null
-         */
-        public $bindParamLengthValue;
     }
 } else {
     /**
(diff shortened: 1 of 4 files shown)


tests/Tracing/Doctrine/DBAL/TracingStatementForV2Test.php
```diff
@@ -137,7 +137,11 @@ public function testBindParamForwardsLengthParamOnlyWhenExplicitlySet(): void
         $this->statement = new TracingStatementForV2($this->hub, $decoratedStatement, 'SELECT 1', ['db.system' => 'sqlite']);
 
         $this->assertTrue($this->statement->bindParam('foo', $variable, ParameterType::INTEGER));
-        $this->assertSame(3, $decoratedStatement->bindParamCallArgsCount);
+        $this->assertSame(4, $decoratedStatement->bindParamCallArgsCount);
+        $this->assertSame(0, $decoratedStatement->bindParamLengthValue);
+
+        $this->assertTrue($this->statement->bindParam('foo', $variable, ParameterType::STRING, 3));
```
ste93cry, line 143: The name of this method is now misleading because it's not true anymore that the `$length` param is forwarded only when explicitly set. Also, all the mess with the custom stub was done because PHPUnit generates the mock so that it always forwards all parameters defined in the method signature to the invocation matcher, regardless of how many you really use in the call. However, since we now forcefully pass `$length` with a proper value, we don't need special handling for it anymore. For this reason, we can just refactorize the `testBindParam()` method so that it uses a data provider where the input test cases are:

- `$length` param is `null` and `$decoratedStatement` receives `0`
- `$length` param is a number and `$decoratedStatement` receives that number

ste93cry (changes requested):
Thank you for taking the time to contribute this change! Besides adding an entry to the CHANGELOG, I just have a doubt about why  the current solution does not work: if you call `bindParam` with up to 3 params, the `array_slice` should return an empty `array`, and the same thing should happen [here](https://github.com/doctrine/dbal/blob/2.13.8/lib/Doctrine/DBAL/Driver/PDOStatement.php#L87). As a consequence of this, PHP should fallback to the default value for the `$length` param 🤔

cjunge-work:
> Thank you for taking the time to contribute this change! Besides adding an entry to the CHANGELOG, I just have a doubt about why the current solution does not work: if you call `bindParam` with up to 3 params, the `array_slice` should return an empty `array`, and the same thing should happen [here](https://github.com/doctrine/dbal/blob/2.13.8/lib/Doctrine/DBAL/Driver/PDOStatement.php#L87). As a consequence of this, PHP should fallback to the default value for the `$length` param thinking

We don't run Sentry on our dev environments, and we have been unable to replicate this issue without Sentry enabled. I'm not sure why the DBAL PDOStatement is working fine but the Sentry TracingStatementv2 isn't.

I haven't done any in-depth exploration, but it appears to me that the default `null` on the tracing statement bindParam method becomes the default if it's not explicitly passed in. This then is passed to the DBALv2 Statement which simply extends PDOStatement. With <PHP8.1 it works fine (I guess PHP implicitly casts it to 0 or ignores it & uses the PDOStatement default), but under PHP8.1 with the changes to deprecate passing null to non-nullable internal function parameters (https://www.php.net/releases/8.1/en.php) the null is triggering the deprecation.

cjunge-work:
@ste93cry I think I've found the smoking gun: the array_slice offset appears to be wrong! https://3v4l.org/7NhQT#v8.1.4

cjunge-work:
@ste93cry I deleted my previous comment as it was wrong. Here's an example showing how the explicit null triggers the deprecation, but not providing the param doesn't: https://3v4l.org/biaIs#v8.1.4. I'm still not sure why the deprecation is being triggered in Sentry.

Jean85:
Can we add a unit test to trigger the deprecation in our CI? Or is it already there, and then why it's not failing?

ste93cry:
>the explicit null triggers the deprecation, but not providing the param doesn't

That's the expected behaviour. If you explicitly pass `null`, then it's up to you to know that you will trigger a deprecation. Silly question: are you sure that there is no other code that pass that param explicitly?

cjunge-work:
@ste93cry My point was that the explicit null triggers the deprecation, when not using Sentry, so it's not a side-effect of the code.

```
<?php

declare(strict_types=1);

use Doctrine\DBAL\Driver\Connection;
use Psr\Container\ContainerInterface;

class aClass
{
    public function __construct(ContainerInterface $container, Connection $connection)
    {
        parent::__construct($container);

        $this->connection = $connection;
    }

    public function aMethod($applicationAverageTime) {
        $statement = $this->connection->prepare('...');
// ...
        $statement->bindParam(':applicationAverageTime', $applicationAverageTime); // <---
// ...
    }
}
```
that's the line that our last deprecation was triggered on, we don't tend to even use the type.

ste93cry:
Sorry for being silent. I reproduced the issue locally, and I found out that the problem originates from [here](https://github.com/doctrine/dbal/blob/dc9b3c3c8592c935a6e590441f9abc0f9eba335b/lib/Doctrine/DBAL/Statement.php#L143). Basically, since the statement is wrapped by our `TracingStatementForV2` class, the `instanceof PDOStatement` returns `false`. Then, the `$length` parameter is explicitly passed to the `bindParam` method with a `null` value and this triggers the deprecation. I would say that the check in the DBAL is pretty weak as-is, and should be changed to take into account the fact that the statement could be wrapped. Anyway, your fix is good, so if you are willing to fix the requested changes I would happily merge this PR.

ste93cry (approved):
I took the time to finish your work, hopefully you don't mind. Thank you for your contribution, I will release the fix in the next days

PR #617, opened 2022-04-15
missing getNativeConnection() method from Connection interface

DeyV:
This method was added in doctrine/dbal 3.3
https://github.com/doctrine/dbal/commit/d9f969b366fcd4c7def89372b9c198edb2cc9fe5

and is required for example in Symfony Message component 

See Doctrine\DBAL\Connection : getNativeConnection()


ste93cry:
Thank you for the contribution. I'm closing this PR because we already introduced support for this new method in #597, but we didn't release the newer version yet

Issue #627, opened 2022-05-16
Call to sentry_trace_meta() breaks with Twig 3.4.0.


jbroutier:
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?
sentry/sentry-symfony 4.2.9

### Steps to Reproduce

1. Setup a Symfony project using sentry/sentry-symfony 4.2+ and twig/twig 3.4.0.
2. Call the sentry_trace_meta() function in a Twig template.

### Expected Result

Template should be rendered without error.

### Actual Result

Symfony returns an exception: "Call to private method Sentry\SentryBundle\Twig\SentryExtension::getTraceMeta() from scope __TwigTemplate_ ..."

Jean85:
This seems very strange to me. We even have a dedicated test for that function: https://github.com/getsentry/sentry-symfony/blob/6177d71f2fd65f4f0f77036a1ca910aa9326767f/tests/Twig/SentryExtensionTest.php#L48-L51

Can you give us additional information?

Jean85:
Relates to https://github.com/twigphp/Twig/issues/3703

PR #628, opened 2022-05-16
fix: Make getTraceMeta() function public.

jbroutier:
Fixes https://github.com/getsentry/sentry-symfony/issues/627

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix compatibility issue with Twig >= 3.4.0 (#628)
+
 ## 4.2.9 (2022-05-03)
 
 - Fix deprecation notice thrown when instrumenting the `PDOStatement::bindParam()` method and passing `$length = null` on DBAL `2.x` (#613)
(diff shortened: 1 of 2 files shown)


Jean85 (changes requested):
On hold until #627 is sorted out.

Jean85 (changes requested):
Fix is fine, can you add a changelog entry?

ste93cry:
Are we really sure that this is not a bug in Twig itself? I mean, it worked fine until now, and suddenly a minor version is released and it breaks everything: it's weird...

Jean85:
It seems related to https://github.com/twigphp/Twig/pull/3702
They changed the way in which they invoke callables, and ours was a private method...

We can try to report it as it is to see if they want to fix it or not.

Jean85 (changes requested):
Stof suggested to us to switch to the more direct `public` method way (so with `[$this, 'methodName]`) since it would be more performant under Twig: https://github.com/twigphp/Twig/issues/3703#issuecomment-1127654569

jbroutier:
PR updated with the direct method call. Do In need to rebase?

PR #635, opened 2022-05-27
SentryTestCommand > Fix deprecated `$defaultName`

ruudk:
This fixes a deprecation on Symfony 6.1:
```
Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Sentry\SentryBundle\Command\SentryTestCommand" class instead.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix compatibility issue with Symfony >= 6.1.0 (#635)
 
 ## 4.2.10 (2022-05-17)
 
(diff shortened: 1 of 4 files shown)


Jean85 (changes requested):
Couldn't we simply add the attribute, so we can avoid both the deprecation and the separation between the command and its name? Attributes are parsed as comments in older PHP versions.

ruudk:
I tried that first (adding AsCommand attribute and keeping the `$defaultName` property) but unfortunately, that gives us this error:
```
The "Symfony\Component\Console\Command\Command::$defaultName" property is considered final. You should not override it in "Sentry\SentryBundle\Command\SentryTestCommand".
```

Jean85 (changes requested):
Ok so the change LGTM.

Can you please add a changelog entry then?

ruudk:
@Jean85 done!

Jean85 (approved):

PR #638, opened 2022-05-31
Tracing: add optionally configurable trace propagation input option to console commands

DAcodedBEAT:
Currently, the "sentry-trace" HTTP header is used for trace propagation,
allowing traces from upstream services to continue traces to downstream services.

In order to trace between the service boundary of a web request to CLI application,
a new optionally configurable CLI option was added to provide the same trace
propagation mechanism, using the same format as the "sentry-trace" header


DAcodedBEAT:
@ste93cry / @Jean85 : According to recent Git history, it looks like you both are the active maintainers for this repository.  I would like to contribute this feature to sentry-symfony so what would be needed to get this reviewed and merged?  Thanks in advance!

ste93cry:
While I certainly find useful what you are trying to achieve, I don't really like the solution. Messing around with all the console commands by adding an input option seems a thing ready to cause errors and headaches in the future. I would prefer to understand if there are other solutions first, but as of now I'm not really inclined to say that I would merge this

DAcodedBEAT:
Thanks for the reply @ste93cry !

I definitely understand the concern around potential future headaches and would feel awful if I caused it somewhere down the line. 😅

What are your thoughts on passing the trace propagation value through an environment variable (perhaps `SENTRY_TRACE` ?) This way, the console command's argv doesn't change as a result of enabling CLI trace propagation.

PR #639, opened 2022-05-31
Start development of version `4.4.x`

ste93cry:


ste93cry:
In the past we used to do it, but a lot of time passed since the last time. I don't remember where, but I read somewhere that regardless of the branch, you should always use `dev-master`

Jean85 (approved):
My bad, I misread the target of the PR.

PR #643, opened 2022-06-21
Drop Symfony support below 4.4

Jean85:
~I'm opening this as a draft since it may require some discussion (or dropping some pieces of code no longer used).~

3.4 is unsupported since last November: https://symfony.com/releases/3.4
Everything between that and 4.4 (since it's an LTS) is unsupported too.

Hence, dropping the support now makes sense. This PR is targeting `develop`, so it would land in the next minor (4.4 as of now).


src/DependencyInjection/SentryExtension.php
```diff
@@ -280,7 +279,6 @@ private function configureErrorListenerIntegration(array $integrations, bool $re
                 [
                     'ignore_exceptions' => [
                         FatalError::class,
-                        FatalErrorException::class,
```
ste93cry, line 283: Are you sure we can stop ignoring this exception? In Symfony `4.4` it is still present

ste93cry (changes requested):
The `RequestListenerTest ::testHandleKernelRequestEvent()` method is still using a data provider that was meant for Symfony < `4.4`

src/DependencyInjection/SentryExtension.php
Jean85: Mh maybe you're right, I'm not sure how 4.4 handles this deprecation. Let's keep it safe, I'll revert this piece.

ste93cry (changes requested):
There are still things to delete: in the `SubRequestListenerTest` class some of the data providers are still splitted per-Symfony version, and probably we can also remove the `KernelEventForwardCompatibilityTrait` trait. I suggest to look for occurences of the `version_compare` or `Kernel::VERSION` keywords

Jean85:
`isMasterRequest` got deprecated in 5.3, `KernelEventForwardCompatibilityTrait` is still needed: [KernelEventForwardCompatibilityTrait](https://github.com/symfony/symfony/blob/a6af0ddb2bf79a8cec20478e4fd8b7bbe83bd1b6/src/Symfony/Component/HttpKernel/Event/KernelEvent.php#L84-L86).

I took care of the rest.

PR #646, opened 2022-07-04
Add `symfony/flex` to the list of Composer plugins allowed to run

ste93cry:
As per title, when we added the `allow-plugins` option to `composer.json` we forgot the `symfony/flex` plugin, which is required to install the same version for all Symfony components in the CI. This PR fixes the issue

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
         "sort-packages": true,
         "allow-plugins": {
             "composer/package-versions-deprecated": true,
-            "phpstan/extension-installer": true
+            "phpstan/extension-installer": true,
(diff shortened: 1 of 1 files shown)


ste93cry:
Apparently, adding the plugin to `allow-plugins` in our `composer.json` is not enough and we must add it to the global one instead because the tool is installed using `shivammathur/setup-php`. I've opened a bug report in shivammathur/setup-php#611 because I think that the GitHub action should do it by itself

ste93cry:
Closing because no more needed. A new release of the GitHub Action fixed the problem 🥳

PR #655, opened 2022-09-23
fix: Update span ops

cleptric:
To unblock the work on Performance Issues, we have to update some span ops.
https://github.com/getsentry/sentry/discussions/37083

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,6 +10,8 @@ parameters:
         - src/aliases.php
     excludePaths:
         - src/aliases.php
+        - src/Tracing/Cache/TraceableCacheAdapterForV2.php
+        - src/Tracing/Cache/TraceableTagAwareCacheAdapterForV2.php
         - src/Tracing/Doctrine/DBAL/TracingStatementForV2.php
         - src/Tracing/Doctrine/DBAL/TracingDriverForV2.php
         - tests/End2End/App
--- a/src/Tracing/Doctrine/DBAL/AbstractTracingStatement.php
+++ b/src/Tracing/Doctrine/DBAL/AbstractTracingStatement.php
@@ -14,7 +14,7 @@ abstract class AbstractTracingStatement
     /**
      * @internal
      */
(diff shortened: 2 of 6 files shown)


tests/Tracing/Twig/TwigTracingExtensionTest.php
```diff
@@ -58,7 +58,7 @@ public function testEnter(Profile $profile, string $spanDescription): void
 
         $this->assertCount(2, $spans);
         $this->assertNull($spans[1]->getEndTimestamp());
-        $this->assertSame('view.render', $spans[1]->getOp());
+        $this->assertSame('symfony.view', $spans[1]->getOp());
```
AbhiPrasad, line 61: Let’s not change this yet - it’s still up for debate

cleptric: https://develop.sentry.dev/sdk/performance/span-operations/#web-server
There is also `template.render`

AbhiPrasad: Yeah - we’ll make decision about this next week and then we can just update it then.

AbhiPrasad (approved):

k-fish (approved):
Looks good!

ste93cry:
I chose the names of these operations because they are what the OpenTelemetry standard uses. Wouldn't it make more sense to support them in Sentry rather than shifting even more from a common standard which is well established in the industry?

AbhiPrasad:
OpenTelemetry doesn’t have the concept of a span op like we do, so fundamentally we’ll be diverging here anyway. Mapping otel span name -> description works for us, but span operation we should be free to work with.

We use the `db` prefix to generate performance issues: https://blog.sentry.io/2022/09/19/performance-issues-slow-you-can-act-on-quickly/

We are working on getting closer to the standard in general though. In https://github.com/getsentry/develop/pull/686 I mapped out the transformation between opentelemetry <-> to sentry data, and we’re working on improving this.

Issue #649, opened 2022-08-01
Improve parameterization of transaction names


smeubank:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which SDK and version?
sentry-symfony

### Steps to Reproduce

similar to https://github.com/getsentry/sentry-laravel/issues/562

### Expected Result

Parameterized transactions as often as possible

### Actual Result

seems symfony only sends raw URLs for everything

smeubank:
we should open a convo with Symfony that they add something to help us access their routing properly

PR #661, opened 2022-10-10
feat: Add support for Dynamic Sampling

cleptric:
This adds support for Dynamic Sampling, introduced in https://github.com/getsentry/sentry-php/pull/1360.


composer.json
```diff
@@ -27,6 +27,7 @@
         "jean85/pretty-package-versions": "^1.5 || ^2.0",
         "php-http/discovery": "^1.11",
         "sentry/sdk": "^3.2",
+        "sentry/sentry": "^3.9",
```
cleptric, line 30: Waiting for https://github.com/getsentry/sentry-php-sdk/pull/15, to be removed.

src/EventListener/TracingRequestListener.php
```diff
@@ -36,9 +37,13 @@ public function handleKernelRequestEvent(RequestEvent $event): void
...
         $requestStartTime = $request->server->get('REQUEST_TIME_FLOAT', microtime(true));
 
-        $context = TransactionContext::fromSentryTrace($request->headers->get('sentry-trace', ''));
+        $context = TransactionContext::fromHeaders(
+            $request->headers->get('sentry-trace', ''),
+            $request->headers->get('baggage', '')
+        );
         $context->setOp('http.server');
         $context->setName(sprintf('%s %s%s%s', $request->getMethod(), $request->getSchemeAndHttpHost(), $request->getBaseUrl(), $request->getPathInfo()));
+        $context->setSource(TransactionSource::route());
```
cleptric, line 46:
```
Suppose this request is instantiated from /mysite on localhost:

 * http://localhost/mysite              returns an empty string
 * http://localhost/mysite/about        returns '/about'
 * http://localhost/mysite/enco%20ded   returns '/enco%20ded'
 * http://localhost/mysite/about?var=1  returns '/about'
```

I'm not entirely sure we can treat this as a `TransactionSource::route()`. To be discussed. Using `TransactionSource::url()` is not an option, as the UX will be terrible.

cleptric:
We made the decision to tackle #649 after the release of DS.

src/Tracing/HttpClient/AbstractTraceableHttpClient.php
```diff
@@ -49,6 +49,7 @@ public function request(string $method, string $url, array $options = []): Respo
         if (null !== $parent) {
             $headers = $options['headers'] ?? [];
             $headers['sentry-trace'] = $parent->toTraceparent();
+            $headers['baggage'] = $parent->toBaggage();
```
cleptric, line 52: ToDo @cleptric Check for `tracePropagationTargets`

composer.json
ste93cry: I think we did this because at the time there was no `symfony / http-client` component available, and thus we decided to stick with the default one provided by the core SDK. I'm more than happy to change this decision, but I wouldn't tackle it here. I suggest to bump `sentry/sdk` to the minimum required version to support DSC and then open a separate PR to remove the dependency and use `sentry/sentry` directly.

ste93cry (approved):

PR #662, opened 2022-10-11
ref: Use constant for the SDK version

cleptric:
Same as on PHP and Laravel. Relying on a constant has fewer side effects and is in line with how we do it on other SDKs.


Jean85 (approved):
LGTM

Is `PrettyVersions` used anywhere in this codebase? Or could we drop it from the direct dependencies?

https://github.com/getsentry/sentry-symfony/blob/a0c76b8026282a3046da47e96139958814c1c88a/composer.json#L27

cleptric:
@Jean85 Still used here https://github.com/getsentry/sentry-symfony/blob/a0c76b8026282a3046da47e96139958814c1c88a/src/DependencyInjection/Configuration.php#L84

PR #675, opened 2022-11-19
Add support for HTTP Client baggage propagation

### LINKED ISSUE #663: Add support for HTTP Client `baggage` propagation
The `AbstractTraceableHttpClient` currently only propagates the `sentry-trace` header.
Add the `baggage` header to outgoing requests as well, if the destination URI is allow listed in `trace_propagation_targets`.

DAcodedBEAT:
I was looking to make use of the `baggage` HTTP header and noticed that this wasn't added to sentry-symfony in yet.

Resolves #663


cleptric:
Thanks for the contribution.
We'll also have to update the `AbstractTraceableHttpClient`

https://github.com/getsentry/sentry-symfony/blob/develop/src/Tracing/HttpClient/AbstractTraceableHttpClient.php#L44-L66

to actually include the `baggage` header on HTTP client requests. The header should only be included if the request's host matches an entry in `trace_propagation_target `.

cleptric (commented):
Please add an assertion for the header in the tests

https://github.com/getsentry/sentry-symfony/blob/ea1a1f5df49ba89350deeaa0455ed2eadedc37a3/tests/Tracing/HttpClient/TraceableHttpClientTest.php#L52

DAcodedBEAT:
Ahh totally missed this change, thanks for pointing out that unit test.  I have updated this in 4321e2a16464f6f2e372e288cde27a98adcd08e9

tests/Tracing/HttpClient/TraceableHttpClientTest.php
```diff
@@ -83,6 +86,102 @@ public function testRequest(): void
         $this->assertSame($expectedTags, $spans[1]->getTags());
     }
 
+    public function testRequestWithInvalidBaggage(): void
```
cleptric, line 89:
```suggestion
    public function testRequestDoesNotContainBaggageHeader(): void
```

tests/Tracing/HttpClient/TraceableHttpClientTest.php
```diff
@@ -83,6 +86,102 @@ public function testRequest(): void
...
+        ];
+
+        $this->assertCount(2, $spans);
+        $this->assertNull($spans[1]->getEndTimestamp());
+        $this->assertSame('http.client', $spans[1]->getOp());
+        $this->assertSame('HTTP PUT', $spans[1]->getDescription());
+        $this->assertSame($expectedTags, $spans[1]->getTags());
+    }
+
+    public function testRequestWithValidBaggage(): void
```
cleptric, line 137:
```suggestion
    public function testRequestDoesContainBaggageHeader(): void
```

PR #677, opened 2022-11-21
feat: Set cache keys as span descriptions

### LINKED ISSUE #619: Unusable trace data called in a callback
### Environment

How do you use Sentry?
Self-hosted (Sentry 21.11.0)

Which SDK and version?
PHP 4.2.8

Used config for sentry in symfony
```
tracing:
        enabled: true
        dbal: # DB queries
            enabled: true
        cache:
            enabled: true
        twig: # templating engine
            enabled: true
```

### Steps to Reproduce

1. Create any kind of a controller for symfony for any kind of a route (POST, GET)
2. Provide a CacheInterface to it (I used redis as an adapter)
3. use code similar to
```php

return $this->cache->get(
            function (ItemInterface $item) {
                $item->expiresAfter($this->cacheTime);

                return $this->handler->handle($someDataWhatever);
            }
        );
```

### Expected Result

Trace in sentry with normal function names. At least $this->handler->handle took 50ms or similar

### Actual Result

image

cleptric:
When possible, set the cache key as the span description when tracing the Symfony cache.

closes #619


michaelzangerle:
Hey @cleptric. I stumbled upon the linked issue and your proposed solution looks nice. Any chance this gets merged soon?

cleptric:
@michaelzangerle Would you mind trying it in one of your apps and reporting back?

michaelzangerle:
@cleptric I just gave it a try and it already looks so much better and clearer in Sentry especially if you have very long list of cache get / save calls.

- Seeing `CLASSMETADATA` helps a lot
- I guess things like this `rApp.Workflow.Config.Message.Response.WorkflowConfigResponse..id` are related to the serializer and I have a lot of those in my traces (maybe there is a way to limit this)
- The only place where it still shows the span id seems to be when `cache.save` is called

image

image

stayallive (approved):

PR #688, opened 2022-12-28
Fix `README.md` badge urls

mhor:
Some badges reference ``sentry/sentry`` package.


stayallive (approved):
Woops, great catch, thanks!

PR #690, opened 2023-01-31
fix: Sanatize HTTP client spans

cleptric:
See https://develop.sentry.dev/sdk/data-handling/#spans & https://develop.sentry.dev/sdk/data-handling/#breadcrumbs.

refs: https://github.com/getsentry/team-webplatform-meta/issues/2


src/Tracing/HttpClient/AbstractTraceableHttpClient.php
```diff
@@ -111,10 +119,4 @@ public function setLogger(LoggerInterface $logger): void
             $this->client->setLogger($logger);
         }
     }
-
-    private function formatUri(Uri $uri): string
-    {
-        // Instead of relying on Uri::__toString, we only use a sub set of the URI
-        return Uri::composeComponents($uri->getScheme(), $uri->getHost(), $uri->getPath(), null, null);
```
cleptric, line 118: The signature is `Uri::composeComponents(?string $scheme, ?string $authority, string $path, ?string $query, ?string $fragment)`, so I feel like this was implemented in a wrong way.

src/Tracing/HttpClient/AbstractTraceableHttpClient.php
```diff
@@ -52,6 +52,12 @@ public function request(string $method, string $url, array $options = []): Respo
             $headers['sentry-trace'] = $parent->toTraceparent();
 
             $uri = new Uri($url);
+            $partialUri = Uri::fromParts([
+                    'scheme' => $uri->getScheme(),
+                    'host' => $uri->getHost(),
+                    'port' => $uri->getPort(),
+                    'path' => $uri->getPath(),
+                ]);
```
stayallive, line 60: CS seems off here:

```suggestion
            $partialUri = Uri::fromParts([
                  'scheme' => $uri->getScheme(),
                  'host' => $uri->getHost(),
                  'port' => $uri->getPort(),
                  'path' => $uri->getPath(),
              ]);
```

And:

`l`: The only "worry" I have here is that a `MalformedUriException` is thrown for some reason. Although this might be totally unfounded since a URI was already there but maybe we want to try/catch anyway just to be 100% sure since not all PSR-7's might be created equal.

src/Tracing/HttpClient/AbstractTraceableHttpClient.php
stayallive: Since it's almost an internal method I think they chose not to bother with default arguments. It's also not part of the PSR-7 spec: https://www.php-fig.org/psr/psr-7/.

stayallive (approved):

PR #709, opened 2023-03-29
Remove advertisement of abandoned package

stayallive:
> https://packagist.org/packages/pugx/sentry-sdk

This package is abandoned because `sentry/sdk` does the exact same now since we switched to using the Symfony HTTP client by default. So we should remove those instructions from the README since they no longer apply.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This is the official Symfony SDK for [Sentry](https://getsentry.com/).
 Using this `sentry-symfony` SDK provides you with the following benefits:
 
  * Quickly integrate and configure Sentry for your Symfony app
- * Out of the box, each event will contain the following data by default 
+ * Out of the box, each event will contain the following data by default
    - The currently authenticated user
    - The Symfony environment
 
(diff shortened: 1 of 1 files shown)


Jean85 (changes requested):
Shouldn't we remove the part above too, that suggests to require an HTTP client manually? Or at least we should update that citing the behavior of the new HTTPlug plugin.

README.md
```diff
@@ -106,38 +106,35 @@ the [PHP specific](https://docs.sentry.io/platforms/php/#php-specific-options) o
...
 ```bash
 composer require php-http/guzzle6-adapter guzzlehttp/psr7
 ```
 
-A possible alternate solution is using `pugx/sentry-sdk`, a metapackage that replaces `sentry/sdk` and uses `symfony/http-client` instead of `guzzlehttp/guzzle`:
-
-```bash
-composer require pugx/sentry-sdk
-```
+If you don't have a compatible HTTP client and/or message factory implementation installed `php-http/discovery` will try to install it for you using a Composer plugin.
```
Jean85, line 130: I would move this before the "If you want to use a different HTTP client" part, do you agree?

Jean85 (approved):

README.md
stayallive: I figured that as long as we keep `sentry/sdk` in here that wouldn't make much sense unless you replace `sentry/sdk` so I figured it would be best to keep it lower until we remove `sentry/sdk` completely since that plugin can be safely ignored as long as we ship with a client out-of-the-box. But I'd be happy to move it higher if you feel that is better!

PR #718, opened 2023-05-03
Replace sentry/sdk by sentry/sentry 3.17

nicolas-grekas:
Same as #693, but updated to 3.17 thanks to https://github.com/getsentry/sentry-php/pull/1504


cleptric:
Holding off on this for the moment. We don't need to fix what is not broken.
We'll revisit this in the next major of the base SDK.

nicolas-grekas:
The thing that is "broken" is sentry/sdk, which forces a specific client implementation, whereas php-http/discovery already does the job. It's super weird that even the recommended solution is to "replace" the SDK.

Linking to https://github.com/getsentry/sentry-symfony/issues/246 for reference.

nicolas-grekas:
Even the README of this package suggests the "replace". This is something that's solved now, let's clear this up :pray:

cleptric:
> The thing that is "broken" is sentry/sdk, which forces a specific client implementation, whereas php-http/discovery already does the job.

We already ran into many issues due to `php-http/discovery`, and right now, I stand by my decision to postpone any further changes in this regard until the next major version.

This is again in strong alignment with our SDK philosophy: https://develop.sentry.dev/sdk/philosophy/#degrade-gracefully.

I wasn't aware of the mentioned passage in our README, I can take a look if this still applies.

nicolas-grekas:
Let's go for a v2?

cleptric:
v4 of https://github.com/getsentry/sentry-php 😄

PR #720, opened 2023-05-11
Improve logging of the user when he logged in programmatically

### LINKED ISSUE #706: User is not set when logged in programatically
Currently the user data is set via https://github.com/getsentry/sentry-symfony/blob/bcdbd3d13376473a514c0ec97566b0c7cdc5a1fc/src/EventListener/RequestListener.php#L66-L75

However, this means that programmatic logins do not add the user data to the event. Instead, an authentication listener should be used so that anywhere a user is logged in, the data is added to the span.

ste93cry:
Fixes #706. I had no time to test it on a real project, so please do it if you can before merging 🙏

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
         "symfony/http-kernel": "^4.4.20||^5.0.11||^6.0",
         "symfony/polyfill-php80": "^1.22",
         "symfony/psr-http-message-bridge": "^1.2||^2.0",
-        "symfony/security-core": "^4.4.20||^5.0.11||^6.0"
+        "symfony/security-core": "^4.4.20||^5.0.11||^6.0",
+        "symfony/security-http": "^4.4.20||^5.0.11||^6.0"
     },
     "require-dev": {
         "doctrine/dbal": "^2.13||^3.0",
@@ -61,7 +62,6 @@
         "symfony/monolog-bundle": "^3.4",
         "symfony/phpunit-bridge": "^5.2.6||^6.0",
         "symfony/process": "^4.4.20||^5.0.11||^6.0",
-        "symfony/security-http": "^4.4.20||^5.0.11||^6.0",
         "symfony/twig-bundle": "^4.4.20||^5.0.11||^6.0",
         "symfony/yaml": "^4.4.20||^5.0.11||^6.0",
         "vimeo/psalm": "^4.3"
--- a/composer.json
+++ b/composer.json
@@ -61,6 +61,7 @@
         "symfony/monolog-bundle": "^3.4",
         "symfony/phpunit-bridge": "^5.2.6||^6.0",
         "symfony/process": "^4.4.20||^5.0.11||^6.0",
+        "symfony/security-http": "^4.4.20||^5.0.11||^6.0",
         "symfony/twig-bundle": "^4.4.20||^5.0.11||^6.0",
         "symfony/yaml": "^4.4.20||^5.0.11||^6.0",
         "vimeo/psalm": "^4.3"
--- a/src/EventListener/LoginListener.php
+++ b/src/EventListener/LoginListener.php
@@ -47,10 +47,6 @@ final class LoginListener
      */
     public function handleAuthenticationSuccessEvent(AuthenticationSuccessEvent $event): void
     {
-        if (class_exists(LoginSuccessEvent::class)) {
-            return;
-        }
-
         $this->updateUserContext($event->getAuthenticationToken());
     }
 
--- a/src/EventListener/LoginListener.php
+++ b/src/EventListener/LoginListener.php
@@ -1,0 +1,129 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\SentryBundle\EventListener;
+
+use Sentry\State\HubInterface;
+use Sentry\State\Scope;
+use Sentry\UserDataBag;
+use Symfony\Component\Security\Core\Authentication\Token\SwitchUserToken;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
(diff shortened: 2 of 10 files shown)


src/EventListener/LoginListener.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * This method is called when an authentication provider authenticates the
+     * user. It is the event closest to {@see LoginSuccessEvent} in versions of
+     * the framework where it doesn't exist.
+     */
+    public function handleAuthenticationSuccessEvent(AuthenticationSuccessEvent $event): void
+    {
+        if (class_exists(LoginSuccessEvent::class)) {
+            return;
```
Jean85, line 51: This line is uncovered. We need a test, or we don't need this check.

Jean85 (dismissed):
Overall a good PR, with a great set of complete and well described tests. Just a few nitpick.

composer.json
```diff
@@ -37,7 +37,8 @@
         "symfony/http-kernel": "^4.4.20||^5.0.11||^6.0",
         "symfony/polyfill-php80": "^1.22",
         "symfony/psr-http-message-bridge": "^1.2||^2.0",
-        "symfony/security-core": "^4.4.20||^5.0.11||^6.0"
+        "symfony/security-core": "^4.4.20||^5.0.11||^6.0",
+        "symfony/security-http": "^4.4.20||^5.0.11||^6.0"
```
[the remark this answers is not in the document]
ste93cry: I agree, definitely. However, making the code optional is a completely different topic that I don't want to tackle here. We need the `security-http` package mostly because of the event classes that are part of it IIRC.

src/EventListener/LoginListener.php
ste93cry: We can remove this check because the event listener is configured dinamically in the `AddLoginListenerTagPass` compiler pass. It was here more as a last-resort safety net rather than for necessity.

ste93cry:
@cleptric can I ask for your review on this?

cleptric:
Yep, can take a look tomorrow/next week.

cleptric (commented):
Do we need to listen to an additional event?

The user ID is only set on the scope during a request that contains credentials, all subsequent requests to a route guarded with `access_control` will only set the user's IP on the scope.

ste93cry:
I found out that the `RequestListener` listener was overwriting the entire user context set on the scope. Since the `LoginListener` probably ran earlier than that, it may have been the cause of seeing only the user IP address for all subsequent requests. I still have some doubts about the fact that we need to listen on the `kernel.request` event to gather the user data from the token storage though because I'm not sure that the `LoginSuccessEvent` event is dispatched on every request...

cleptric:
Both `LoginListener::handleLoginSuccessEvent` and `LoginListener::handleAuthenticationSuccessEvent` seem not to be called on each request. In my controller, `$this->security->getUser()` does return the logged-in user. The event sent to Sentry only contains the IP.

Screenshot 2023-07-25 at 22 12 35

ste93cry:
Do you mind sharing the `security.yaml` file you're using in your test project? In mine, I configured it like this and it seems to work fine:

```yaml
security:
    password_hashers:
        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'plaintext'

    providers:
        users_in_memory:
            memory:
                users:
                    john: { password: 'doe', roles: ['ROLE_USER'] }
    firewalls:
        main:
            provider: users_in_memory
            http_basic:
                realm: Secured Area

    access_control:
        - { path: ^/, roles: ROLE_USER }
```

cleptric:
```yaml
security:
    # https://symfony.com/doc/current/security.html#registering-the-user-hashing-passwords
    password_hashers:
        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
    # https://symfony.com/doc/current/security.html#loading-the-user-the-user-provider
    providers:
        # used to reload user from session & other features (e.g. switch_user)
        app_user_provider:
            entity:
                class: App\Entity\User
                property: email
    firewalls:
        dev:
            pattern: ^/(_(profiler|wdt)|css|images|js)/
            security: false
        main:
            lazy: true
            provider: app_user_provider

            form_login:
                # "app_login" is the name of the route created previously
                login_path: login
                check_path: login

            logout:
                path: logout
                # where to redirect after logout
                target: /

            # activate different ways to authenticate
            # https://symfony.com/doc/current/security.html#the-firewall

            # https://symfony.com/doc/current/security/impersonating_user.html
            # switch_user: true

    # Easy way to control access for large sections of your site
    # Note: Only the *first* access control that matches will be used
    access_control:
        - { path: ^/sentry, roles: ROLE_USER }

when@test:
    security:
        password_hashers:
            # By default, password hashers are resource intensive and take time. This is
            # important to generate secure password hashes. In tests however, secure hashes
            # are not important, waste resources and increase test times. The following
            # reduces the work factor to the lowest possible values.
            Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface:
                algorithm: auto
                cost: 4 # Lowest possible value for bcrypt
                time_cost: 3 # Lowest possible value for argon
                memory_cost: 10 # Lowest possible value for argon
```

cleptric:
https://github.com/cleptric/symfony - Most logic is in the `SentryController`.

ste93cry:
I have no clue on why it was working in my test project, but indeed the framework does not dispatch the two login-related events for each request. I readded back a listener for the `request.kernel` event and it now works as expected.

cleptric:
Screenshot 2023-07-27 at 13 14 16

Looking good!

@Jean85 can you take another look, please?

cleptric (approved):

PR #721, opened 2023-05-22
Fix wrong type of http_connect_timeout|http_timeout options

ostrolucky:
It has to be float, otherwise it's not possible to set it lower than 1s

See: https://github.com/symfony/symfony/blob/8e7ea19541b70aab3206a681dee497813f7b02f2/src/Symfony/Contracts/HttpClient/HttpClientInterface.php#L55-L56


cleptric (approved):
Thanks!

PR #722, opened 2023-05-25
Minor: Use connection instead connecton in phpdoc

alamirault:
`connection` is always used but not in this phpdoc


cleptric (approved):
Cool, thanks! 🚀

Issue #730, opened 2023-06-16
4.9.0 version doesn't load the correct DBAL Platform


jdecool:
### Environment

- PHP 8.1.20
- Symfony 5.4
- DBAL 3.6.3 / ORM 2.14.3
- Sentry Symfony 4.9.0

### Steps to Reproduce

On a project I work, I've updated `sentry/sentry-symfony` from `4.8.0` to `4.9.0`.

After the update I've this exception when trying to interact with the database:

```text
In AbstractPlatform.php line 441:
                                                                               
  Unknown database type json requested, Doctrine\DBAL\Platforms\MySQLPlatform  
   may not support it.  
```

It seems that https://github.com/getsentry/sentry-symfony/pull/723 change the `AbstractPlaform` loaded instance.
In my configuration I specify the `serverVersion` in the Doctrine DSN.

With the `4.8.0` I've the `Doctrine\DBAL\Platforms\MySQL80Platform` which is loaded whereas when I update dependency to `4.9.0` load `Doctrine\DBAL\Platforms\MySQLPlatform` which not implement the `json` type mapping.

ste93cry:
I bet the problem is that we're not implementing the `createDatabasePlatformForVersion()` method in the driver's `class`. As the fix seems rather easy, would you mind opening a PR?

jdecool:
Sure.

I've test and it fix the bug.

I'm going to create the PR.

jdecool:
I've just created https://github.com/getsentry/sentry-symfony/pull/731

kov-lucas:
I think this is also breaking the `bin/console doctrine:migrations:diff` command, this is the only dependency that i've recently updated and now it is completely broken, i had to roll it back to 4.8.0

cleptric:
Version [4.9.1](https://github.com/getsentry/sentry-symfony/releases/tag/4.9.1) was released which contains a fix for this issue.

PR #731, opened 2023-06-16
Make TracingDriverForV32 implements VersionAwarePlatformDriver to handle platform version

jdecool:
Fix https://github.com/getsentry/sentry-symfony/issues/730

As discussed in the issue, I had the `createDatabasePlatformForVersion` method to handle the platform version.

### DIFF
--- a/tests/Tracing/Doctrine/DBAL/TracingDriverForV32Test.php
+++ b/tests/Tracing/Doctrine/DBAL/TracingDriverForV32Test.php
@@ -109,4 +109,40 @@ final class TracingDriverForV32Test extends DoctrineTestCase
 
         $this->assertSame($exceptionConverter, $driver->getExceptionConverter());
     }
+
+    public function testCreateDatabasePlatform(): void
+    {
+        $databasePlatform = $this->createMock(AbstractPlatform::class);
+
+        $decoratedDriver = $this->createMock(DriverInterface::class);
+        $decoratedDriver->expects($this->once())
+            ->method('getDatabasePlatform')
+            ->willReturn($databasePlatform);
+
+        $driver = new TracingDriverForV32($this->connectionFactory, $decoratedDriver);
+
+        $this->assertSame($databasePlatform, $driver->createDatabasePlatformForVersion('5.7'));
+    }
+
+    public function testCreateDatabasePlatformForVersionWhenDriverDefinedCreateDatabasePlatformForVersion(): void
+    {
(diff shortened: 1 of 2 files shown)


ste93cry:
We need to implement the method without implementing the interface, otherwise we trigger again the deprecation. Please also add some tests for the new method 🙏

ste93cry:
>Do you prefer I've implement the test in an other way ?

Since the whole point of this class is to decouple from the `VersionAwarePlatformDriver` deprecated interface, I don't think we should use it in the tests either. What about mocking the `Driver` interface using `getMockBuilder()` and then using `addMethods()` on it to configure the `createDatabasePlatformForVersion()` method? Or, we could create a fake interface with that method declared and then simply use `createMock()` on it, similar to what we did in [`TracingDriverForV2`](https://github.com/getsentry/sentry-symfony/blob/707212c4330ccf2d2790f0f0a7036f4ace66fa7b/tests/Tracing/Doctrine/DBAL/TracingDriverForV2Test.php#L192)

jdecool:
I choose to create a fake interface.

tests/Tracing/Doctrine/DBAL/TracingDriverForV32Test.php
```diff
@@ -109,4 +109,45 @@ public function testGetExceptionConverter(): void
...
+
+if (interface_exists(DriverInterface::class)) {
+    interface StubCreateDatabasePlatformForVersionDriver extends DriverInterface
+    {
+        /**
+         * @param string $version
+         *
+         * @return AbstractPlatform
+         */
+        public function createDatabasePlatformForVersion($version);
```
ste93cry, line 151: Given that the method does not exist in the parent interface and when it will be added it will be typehinted anyway, it makes more sense to avoid the PHPDoc

```suggestion
        public function createDatabasePlatformForVersion(string $version): AbstractPlatform;
```

ste93cry (changes requested):
Apart from a small nitpick, it looks good to me. Do you mind testing that it works well in your project and confirm it?

tests/Tracing/Doctrine/DBAL/TracingDriverForV32Test.php
jdecool: ✅

ste93cry (approved):
Good work, thank you!

jdecool:
Thanks @ste93cry for your help 🙏🏼

PR #750, opened 2023-07-31
Add support for `symfony/psr-http-message-bridge: ^6.4`

### LINKED ISSUE #749: Your requirements could not be resolved to an installable set of packages.
### How do you use Sentry?

Self-hosted / on-premises

### SDK version

4.9.2

### Steps to reproduce

composer update -W

### Expected result

Problem 1
    - Root composer.json requires sentry/sentry-symfony ^4.9.2 -> satisfiable by sentry/sentry-symfony[4.9.2].
    - sentry/sentry-symfony 4.9.2 requires symfony/psr-http-message-bridge ^1.2||^2.0 -> found symfony/psr-http-message-bridge[v1.2.0, v1.3.0, v2.0.0, ..., v2.3.1] but these were not loaded, likely because it conflicts with another require.

### Actual result

Now I have installed symfony/psr-http-message-bridge v2.2.0.

cleptric:
Hopefully fixes #749 once Symfony 6.4 was tagged.


javiereguiluz:
I can confirm that this change works. I updated a Symfony app to 6.4 and used my own fork of `sentry/sentry-symfony` that contained this exact change. Thanks.

fabpot:
Just tested as well and it works for me.

composer.json
```diff
@@ -36,7 +36,7 @@
         "symfony/event-dispatcher": "^4.4.20||^5.0.11||^6.0",
         "symfony/http-kernel": "^4.4.20||^5.0.11||^6.0",
         "symfony/polyfill-php80": "^1.22",
-        "symfony/psr-http-message-bridge": "^1.2||^2.0",
+        "symfony/psr-http-message-bridge": "^1.2||^2.0||^6.4",
```
stof, line 39:
```suggestion
        "symfony/psr-http-message-bridge": "^1.2||^2.0||^6.4||^7.0",
```

so that this supports also Symfony 7 (which will be released at the same date than 6.4)

cleptric: Supporting Smyonfy 7 is more involved than this, and we already have a PR open. See https://github.com/getsentry/sentry-symfony/pull/761.
I will focus this PR on 6.4 only.

nicolas-grekas:
The root issue should be fixed by https://github.com/symfony/flex/pull/1003, but this change is still welcome (I'd suggest merging + releasing without waiting for 6.4-stable, that might make things a big smoother for end users.)

ste93cry (approved):

cleptric:
Released in [4.12.0](https://github.com/getsentry/sentry-symfony/releases/tag/4.12.0)

PR #753, opened 2023-08-19
Prefer the `SENTRY_RELEASE` environment variable over the package root version

### LINKED ISSUE #623: Precedence order when setting app release
When working with the Symfony SDK I noticed that the app version set in composer.json as below is prioritised over the SENTRT_RELEASE environment variable set in .env file. 

```
{
    "type": "project",
    "license": "proprietary",
    "version": "1.58.1",
    "minimum-stability": "stable",
}
```

The precedence order according to my tests was: 

1. Release set in config/packages/sentry.yaml
2. “Version” tag added to composer.json
3. SENTRY_RELEASE environment variable set in .env

Because only the environment variable and `sentry.yaml` options are documented, I would expect these two to take precedence over all other automatically detected releases. Is this intended behaviour? 

I’m using:
Sentry Symfony SDK v4.2.9
Composer v2.0.9
Symfony v5.4.8 
PHP v7.4.29

ste93cry:
Fixes #623. Since the default value for the `release` option is hardcoded to be the root package version, the `SENTRY_RELEASE` environment variable is never considered unless explicitly referenced in the config by the user. With these changes, the environment variable is used as default value. If its value is an empty string or the variable is not set, the root package version is used as a fallback.

This is technically speaking a change in the behaviour of the config, so it may be worth to document it. Also, this may be considered a bugfix rather than an improvement since it was never intended to prefer the root package version over the environment variable.

### DIFF
--- a/src/DependencyInjection/SentryExtension.php
+++ b/src/DependencyInjection/SentryExtension.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Sentry\SentryBundle\DependencyInjection;
 
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
+use Jean85\PrettyVersions;
 use Psr\Log\NullLogger;
 use Sentry\Client;
 use Sentry\ClientBuilder;
@@ -66,6 +67,8 @@ final class SentryExtension extends ConfigurableExtension
     {
         $loader = new Loader\XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('services.xml');
+
+        $container->setParameter('env(SENTRY_RELEASE)', PrettyVersions::getRootPackageVersion()->getPrettyVersion());
 
         $this->registerConfiguration($container, $mergedConfig);
         $this->registerErrorListenerConfiguration($container, $mergedConfig);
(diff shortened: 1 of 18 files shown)


src/DependencyInjection/SentryExtension.php
```diff
@@ -67,6 +68,8 @@ protected function loadInternal(array $mergedConfig, ContainerBuilder $container
         $loader = new Loader\XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('services.xml');
 
+        $container->setParameter('env(SENTRY_RELEASE)', PrettyVersions::getRootPackageVersion()->getPrettyVersion());
+
```
Jean85, line 72: Wouldn't this risk the reverse, overriding the env-provided value, depending on the order of configuration loading?

Jean85 (commented):
This env behavior was already documented like this, so I would consider it as a bugfix. But, this wasn't working because the bundle is overriding the default SDK behavior here: https://github.com/getsentry/sentry-php/blob/95d3d3a518010299fc928609ddb05f0067f0b6ac/src/Options.php#L906C1-L906C61

IMHO we should just reverse the priority, and it seems that this PR is doing that... I have a doubt though...

src/DependencyInjection/SentryExtension.php
ste93cry: I agree that it is better to play safe and set the parameter only if it doesn't exist yet. I also refactorized the test case for the fallback to the env var to use a different approach, so now it resembles more closely what would happen by setting an environment variable for real.

Jean85 (approved):

PR #761, opened 2023-09-13
Symfony 7 support

Jean85:
This is a playground to act upon https://github.com/symfony/symfony/issues/51563

Symfony 7 is coming this November, and it will include a type overhaul that will impact the ecosystem, and us too.

In this PR, I tried forcibly upgrading to it, and 968b232 lists all the changes that we need to do; I'll comment them accordingly.

Unfortunately, at least the cache adapters require a `mixed` return type, which is a breaking change. We would need to either duplicate and selectively autoload those classes (as we already did with DBAL v2 vs v3) or do a breaking change, [requiring PHP 8.0 as a minimum](https://caniphp.com/?s=mixed).


src/Tracing/Cache/TraceableCacheAdapterForV2.php
```diff
@@ -40,7 +40,7 @@ public function __construct(HubInterface $hub, AdapterInterface $decoratedAdapte
      *
      * @return mixed
      */
-    public function get(string $key, callable $callback, float $beta = null, array &$metadata = null)
+    public function get(string $key, callable $callback, float $beta = null, array &$metadata = null): mixed
```
Jean85, line 43: This would be a breaking change. 
Selective autoload could be a viable, but messy solution.

src/Tracing/Cache/TraceableTagAwareCacheAdapterForV2.php
```diff
@@ -41,7 +41,7 @@ public function __construct(HubInterface $hub, TagAwareAdapterInterface $decorat
      *
      * @return mixed
      */
-    public function get(string $key, callable $callback, float $beta = null, array &$metadata = null)
+    public function get(string $key, callable $callback, float $beta = null, array &$metadata = null): mixed
```
Jean85, line 44: Same here: this would be a breaking change.
Selective autoload could be a viable, but messy solution.

tests/EventListener/Fixtures/UserWithoutIdentifierStub.php
```diff
@@ -31,4 +31,9 @@ public function getSalt(): ?string
     public function eraseCredentials(): void
     {
     }
+
+    public function getUserIdentifier(): string
+    {
+        return $this->getUsername();
+    }
```
Jean85, line 38: This is a non breaking change, and it's on a test stub. Needed, but not an issue.

src/Tracing/Cache/TraceableCacheAdapterForV2.php
ste93cry: We already have this return type in `TraceableCacheAdapterForV3` 🤔

Jean85: Oh wow you're right! It's probably due to the PSR evolution... But this would mean that Symfony 7 + DBAL 2 is a no-go.

If that's already the case outside of us, that would mean that we're already good as we are!

ste93cry: I don't get what the DBAL has to do with this. Do you mind explaining?

Jean85: Wops sorry, my mistake. This V2 in the name is referring to PSR-16, through Symfony Constracts, right?

ste93cry: Yes, that's correct!

mbabker: > But this would mean that Symfony 7 + DBAL 2 is a no-go.

I know the conversation moved past this, but https://github.com/symfony/symfony/pull/50571 did drop DBAL 2 support in Symfony 7.

tests/EventListener/Fixtures/UserWithoutIdentifierStub.php
ste93cry: Why is this needed? The stub is named `UserWithoutIdentifierStub` to indicate that it doesn't implement the newer `getUserIdentifier()` method, which is something that we definitely need to test on older versions of the framework

src/Tracing/Cache/TraceableTagAwareCacheAdapterForV2.php
ste93cry: Same as above, please revert this change as it's not needed. We have `TraceableTagAwareCacheAdapterForV3` specifically for newer Symfony versions that changed the signature of this method

tests/EventListener/Fixtures/UserWithoutIdentifierStub.php
stof: what is probably needed is to skip the test using that stub when running tests against a version of Symfony where `getUserIdentifier` is part of the interface.

ste93cry: This should already be the case, which is why I don't understand this change. @Jean85 did you do it on purpose because something was broken or just because such a method is required in new versions of the framework?

Jean85: IIRC it was needed to test a codepath that is triggered against older versions of Symfony which do NOT have that method in the interface.

composer.json
```diff
@@ -30,15 +30,17 @@
...
-        "symfony/event-dispatcher": "^4.4.20||^5.0.11||^6.0",
-        "symfony/http-kernel": "^4.4.20||^5.0.11||^6.0",
+        "symfony/config": "6.4.x-dev as 6.4.99",
+        "symfony/console": "6.4.x-dev as 6.4.99",
+        "symfony/dependency-injection": "6.4.x-dev as 6.4.99",
+        "symfony/event-dispatcher": "6.4.x-dev as 6.4.99",
+        "symfony/http-foundation": "6.4.x-dev as 6.4.99",
+        "symfony/http-kernel": "6.4.x-dev as 6.4.99",
         "symfony/polyfill-php80": "^1.22",
         "symfony/psr-http-message-bridge": "^1.2||^2.0",
```
cleptric, line 40:
```suggestion
        "symfony/psr-http-message-bridge": "^1.2||^2.0||^7.0",
```

stof:
what is the goal of this `Force Symfony 6.4-dev` commit ?

ste93cry:
I don't know, but I reverted the changes and added Symfony `7.*` to the test matrix

javiereguiluz:
About this concern from the original PR description:

> "Unfortunately, at least the cache adapters require a mixed return type, which is a breaking change. We would need to either [...] or do a breaking change, requiring PHP 8.0 as a minimum."

FYI, active support for PHP 8.0 ended a year ago and security support will end in 19 days. Also, active support for PHP 8.1 ends in 18 days. See https://www.php.net/supported-versions.php

So, it's probably OK to change the PHP requirements. Thanks!

cleptric:
No, it's not ok to raise the PHP requirement 🙂 

~17% of our users are still below PHP 8.0 https://packagist.org/packages/sentry/sentry-symfony/php-stats#4.

janklan:
> ~17% of our users are still below PHP 8.0

Isn't a new major version release a solution to introducing bc breaks? Why not release a 5.x and let the sloths be sloths?

cleptric:
> > ~17% of our users are still below PHP 8.0
> 
> Isn't a new major version release a solution to introducing bc breaks? Why not release a 5.x and let the sloths be sloths?

Because this is not how Sentry operates, have a look at https://develop.sentry.dev/sdk/philosophy/#compatibility-is-king.

janklan:
> Because this is not how Sentry operates, have a look at https://develop.sentry.dev/sdk/philosophy/#compatibility-is-king.

Yeah, I get that, but releasing a new major version with breaking changes doesn't prevent anyone from maintaining the 4.x branch with legacy code support.

Either way, I assume you guys will work something out to support Symfony 7. Compatibility is king, like you said.

composer.json
```diff
@@ -30,15 +30,15 @@
         "sentry/sdk": "^3.4",
```
evertharmeling, line 30: We should also change this to `"sentry/sdk": "^3.4||^4.0"` as version `3.4` is bound to `symfony/http-client ^4.3|^5.0|^6.0`

ste93cry: There's an open [PR](https://github.com/getsentry/sentry-php-sdk/pull/22) that adds support for Symfony `7.x` to `sentry/sdk:3.x`. We cannot support `sentry/sdk:4.0` yet because it would allow installing the newest SDK, which introduced several breaking changes

cleptric: Nope, this is a BC change. `sentry/sdk: ^3.4` will also gain Symonfy 7 support, no worries.

janklan:
You're on fire, @ste93cry!

Jean85:
We're getting failures due to deprecations on a `--prefer-lowest` job, maybe we should silence deprecations in that case?

ste93cry:
>we should silence deprecations in that case?

There's no need to, I reinstated the dependency on the `symfony/dom-crawler` package that I removed earlier, without realizing that it was required explicitly to avoid those deprecations.

ste93cry (approved):
>we should silence deprecations in that case?

There's no need to, I reinstated the dependency on the `symfony/dom-crawler` package that I removed earlier, without realizing that it was required explicitly to avoid those deprecations. Changes look good now, CI is finally 🟢, so LGTM for me

cleptric (approved):
Thanks @Jean85 & @ste93cry!

PR #767, opened 2023-10-22
Add jobs for PHP `8.3` to the CI

ste93cry:
With PHP `8.3` coming in November 2023, it's time to run the CI for that version as well so that we're ready for its release.

### DIFF
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.3'
+          php-version: '8.2'
 
       - name: Install dependencies
         run: composer update --no-progress --no-interaction --prefer-dist
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.1'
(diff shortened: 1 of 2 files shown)


.github/workflows/static-analysis.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.1'
+          php-version: '8.3'
```
Jean85, line 22: This is failing, and PHP-CS-Fixer is slow to support newer language versions. We should stick to 8.2 here
```suggestion
          php-version: '8.2'
```

Jean85 (changes requested):

.github/workflows/static-analysis.yaml
ste93cry: >PHP-CS-Fixer is slow to support newer language versions

It makes me sad, but I agree. I wanted to run the tol on PHP `8.3` just for the sake of consistency with the other jobs, but I remember that in the past we had to wait on average a long time for a compatible version to be released

ste93cry:
@cleptric I think it's pointless to retest all PHP versions just to make sure they work without optional packages installed, so I removed some jobs from the GitHub Workflow and changed the matrix so that only the oldest and newest PHP versions of each major are tested. However, as you can see from the following image, some jobs are still flagged as required checks for the CI to pass. Maybe the repo settings need to be adjusted accordingly?

immagine

cleptric:
The required checks haven been removed.

Jean85 (approved):

PR #768, opened 2023-10-23
Update symfony/psr-http-message-bridge dependency

codedmonkey:
The Symfony team is preparing the [psr-http-message-bridge](https://github.com/symfony/psr-http-message-bridge) to be included in the main Symfony release schedule, moving it from the current version 2.3 to version 6.4 and 7.0 simultaneously. This is going to cause issues for developers using Symfony as described in my issue [here](https://github.com/symfony/flex/issues/998).

To fix this, I've added the `^6.4` constraint for the package in composer.json.


ste93cry:
Thanks for submitting this PR, but I'm closing it because it's a duplicate of #750

PR #775, opened 2023-10-30
Prepare 4.12.0

cleptric:


ste93cry:
Should we mention that we now test for compatibility with PHP `8.3`?

cleptric:
I normally omit this internal stuff, as it has little impact on our users. Like, I'm expecting a library to test their code on the upcoming PHP version at one point prior release by default.

ste93cry (approved):
Then, LGTM ✅

PR #785, opened 2023-11-29
Remove monolog dependency

Melfo01:
Monolog repository is not used directly by this repository; so we can remove it ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,6 @@
         "doctrine/dbal": "^2.13||^3.0",
         "doctrine/doctrine-bundle": "^1.12||^2.5",
         "friendsofphp/php-cs-fixer": "^2.19||<=3.16.0",
-        "monolog/monolog": "^1.3||^2.0||^3.0",
         "phpstan/extension-installer": "^1.0",
         "phpstan/phpstan": "^1.3",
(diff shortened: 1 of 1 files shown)


Melfo01:
Hey @cleptric 👋,

My tests fail with Symfony 6.0. I think the problem is caused by `symfony/psr-http-message-bridge` package. In master, tests was launched with version 2.3.1, my PR used 6.4 version. 

Do you know why and how to fix it?
Same question with PHPStan?

Jean85 (approved):

ste93cry:
I have nothing against merging these changes, but I fail to understand how it could help with your problem as `monolog/monolog` is a development requirement and does not affect at all third parties.

Jean85:
Well, at the very least we're testing that it works fine with that major.

ste93cry:
Technically speaking, we don't even depend on `monolog/monolog` in this bundle, but rather on `symfony/monolog-bundle`, which is needed only for the E2E tests anyway. We could even remove the requirement on the lib imho, I don't see why it would be useful to keep it.

ste93cry:
Yes please, that would be cool!

ste93cry (approved):

PR #786, opened 2023-11-29
Fix the decoration of the HTTP client when tracing is enabled

### LINKED ISSUE #700: TraceableHttpClientForV6 twice in symfony/http-client decoration chain

ste93cry:
The distributed tracing functionality for the HTTP client has been broken for a long time: first, the decoration of the raw HTTP client does not happen correctly, leading to a double presence of our instrumentation in the decoration chain. Second, but not least, because our instrumentation runs before we have the complete URL of the request, the reported span includes incomplete and/or misleading information.

This PR is an attempt to fix #700 based on @simonberger's contribution in #704. Starting from Symfony `6.3`, the framework decoration strategy was revisited and this made the fix quite simple. However, on older versions of the framework, each built-in decorator of the HTTP client (e.g. the `RetryableHttpClient`), references in the constructor the `.inner` service of the client being decorated, making impossible to inject our own client at a given position because no real decoration chain exists. For this reason, I had to hardcode and decorate the specific services created by the framework in [`FrameworkExtension`](https://github.com/symfony/symfony/blob/ce95b8763a168756e4e050a39ae7ca5258a2bde0/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2334), rather than the raw HTTP client.

### DIFF
--- a/src/DependencyInjection/Compiler/HttpClientTracingPass.php
+++ b/src/DependencyInjection/Compiler/HttpClientTracingPass.php
@@ -13,6 +13,17 @@ use Symfony\Component\DependencyInjection\Reference;
 final class HttpClientTracingPass implements CompilerPassInterface
 {
     /**
+     * List of service IDs that can be registered in the container by the
+     * framework when decorating the mock client. The order is from the
+     * outermost decorator to the innermost.
+     */
+    private const MOCK_HTTP_CLIENT_SERVICE_IDS = [
+        'http_client.retryable.inner.mock_client',
+        '.debug.http_client.inner.mock_client',
+        'http_client.mock_client',
+    ];
+
+    /**
      * {@inheritdoc}
      */
     public function process(ContainerBuilder $container): void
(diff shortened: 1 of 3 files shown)


simonberger:
It is working correctly in my tests on Symfony 5.4. Doing it in the compile stage makes detection easier, good job. I guess I did not make it to work there, because I forgot to re-set the decoration priority when switching the decoration class.
Approving from my side.

PR #789, opened 2023-12-03
Allow sentry/sdk 4

tacman:
A tiny step toward #761


Jean85:
I'm sorry but this is not feasible at all. The new SDK has numerous breaking changes, and it's not possible to do i that way. I'll approve the CI to show how much this is true.

tacman:
Understood.  I thought it was necessary for Symfony 7, but really version 3 of the SDK shouldn't block Symfony 7 or PHP 8.3.  

Thx.

PR #797, opened 2024-01-05
fix: http client aliases

### LINKED ISSUE #794: Add conflicting package symfony/http-client v6
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.13.1

### Steps to reproduce

Run an application with `sentry/sentry-symfony` in combination with `symfony/http-client-contracts` on version 3.x, which can be achieved by using `symfony/http-client` on version 6 or higher.

### Expected result

No errors

### Actual result

I was working on updating one of our applications to PHP 8.2, including a number of packages. The following error appeared running the automated tests:

```
PHP Fatal error:  Declaration of Sentry\SentryBundle\Tracing\HttpClient\TraceableHttpClientForV5::withOptions(array $options): Sentry\SentryBundle\Tracing\HttpClient\TraceableHttpClientForV5 must be compatible with Symfony\Contracts\HttpClient\HttpClientInterface::withOptions(array $options): static in /home/mschwartz/Documents/projects/testjeleefstijl/server/vendor/sentry/sentry-symfony/src/Tracing/HttpClient/TraceableHttpClientForV5.php on line 15
```

Turns out `TraceableHttpClientForV5::withOptions()` is defined as `public function withOptions(array $options): self`, whereas [the parent](https://github.com/symfony/http-client-contracts/blob/main/HttpClientInterface.php#L98) defines this as `public function withOptions(array $options): static`

I solved this adding `"symfony/http-client": "^5.4",` to our `composer.json`, because it was implicitly using version 6, which requires `symfony/http-client-contracts` version 3.

My suggestion is to add `"symfony/http-client": "^6|^7",` as conflicting packages.

victor-prdh:
Hello,

Attempt to fix #794.

Since theres is no specific changes from http-client 5.4 to 6.0 ([see changelog](https://github.com/symfony/symfony/blob/v6.0.0/src/Symfony/Component/HttpClient/CHANGELOG.md)), there is no way to know the version with the current way of doing things (i.e. checking if class / interface / methods exist or not) i propose to use `InstalledVersions::getVersion` of composer.

I saw the `jean85/pretty-package-versions` package but it's "overkill" in this specific case IMO (we just need the package version, not an object to deal with)

Thanks for your review !

### DIFF
--- a/src/aliases.php
+++ b/src/aliases.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sentry\SentryBundle;
 
+use Composer\InstalledVersions;
 use Doctrine\DBAL\Result;
 use Sentry\SentryBundle\Tracing\Cache\TraceableCacheAdapter;
 use Sentry\SentryBundle\Tracing\Cache\TraceableCacheAdapterForV2;
@@ -27,7 +28,6 @@ use Sentry\SentryBundle\Tracing\HttpClient\TraceableResponseForV5;
 use Sentry\SentryBundle\Tracing\HttpClient\TraceableResponseForV6;
 use Symfony\Component\Cache\Adapter\AdapterInterface;
 use Symfony\Component\Cache\DoctrineProvider;
-use Symfony\Component\HttpClient\Response\StreamableInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
 if (interface_exists(AdapterInterface::class)) {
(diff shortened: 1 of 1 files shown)


Jean85 (commented):
This may work, but it may break on edge cases, such installing a fork of the original package or using `symfony/symfony`, which may break the `getVersion` call due to package replacement.

victor-prdh:
Mmh, didn't think about that...
About using `symfony/symfony` we can imagine doing that can handle it
`\Composer\InstalledVersions::getVersion('symfony/http-client') ?? \Composer\InstalledVersions::getVersion('symfony/symfony')`.

In a fork case, unfortunately i have no idea on how we can handle it...

Jean85:
The null coalesce is not enouth, IIRC the method throws in case there's a replaced package!

Jean85:
@victor-prdh can you rework this following this suggestion?

https://github.com/getsentry/sentry-symfony/issues/794#issuecomment-1881201723

> The existence of `HttpClientInterface::stream` and `HttpClientInterface::withOptions` seems to be the path to follow!

victor-prdh:
Thanks for this suggestion @Jean85 !

Jean85 (approved):
LGTM! :+1: 

#799 should fix the build.

ste93cry:
I just wonder why we changed the `interface_exists` check from `Symfony\Contracts\HttpClient\\ResponseInterface` to `Symfony\Contracts\HttpClient\\HttpClientInterface`. Besides that, thanks for contributing!

victor-prdh:
We now need to check a method existence on `HttpClientInterface`. This interface was already here in Symfony 4.4 so as we must use it after, we can  save an use of `ResponseInterface` and use only `HttpClientInterface`. Not a big optimisation but why not

ste93cry:
Thanks, that makes totally sense!

PR #808, opened 2024-02-08
Change the way the Doctrine tracing middleware is registered in DBAL `>=3.x`

### LINKED ISSUE #805: DbalTracingPass is overridden by DoctrineBundle's own MiddlewaresPass
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.13.2

#### Relevant dependencies:

| Package | Version |
|----------|--------|
| symfony  | 5.4 |
| symfony/dependency-injection  | 5.4.34 |
| doctrine/doctrine-bundle | 2.11.1  |
| doctrine/dbal | 3.7.3 |

Nothing special otherwise, just have Sentry-Symfony and its profiling integration activated.

### Expected result

Sentry hooks into DBAL, to enable its profiling functionality for queries

### Actual result

Sentry fails to hook into DBAL, because DBALs `MiddlewarePass` adds another setMiddlewares invocation that ends up overwriting the call placed there by `DbalTracingPass`. `MiddlewarePass` always seems to place that call, no matter if the set of middlewares it adds is empty or not (e.g. with or without Symfony's own profiling middleware).

I suppose the order in which these passes run is important (it'll work if DbalTracingPass runs last (because it removes a previous setMiddlewares call anyway), but it won't work the other way around)

Looking at the generated container in var/cache:

```php
protected function getDoctrine_Dbal_DefaultConnectionService()
    {
        $a = new \Doctrine\DBAL\Configuration();

        $b = new \Doctrine\Bundle\DoctrineBundle\Middleware\DebugMiddleware(($this->privates['doctrine.debug_data_holder'] ?? ($this->privates['doctrine.debug_data_holder'] = new \Doctrine\Bundle\DoctrineBundle\Middleware\BacktraceDebugDataHolder([]))), ($this->privates['debug.stopwatch'] ?? ($this->privates['debug.stopwatch'] = new \Symfony\Component\Stopwatch\Stopwatch(true))));
        $b->setConnectionName('default');

        $a->setSchemaManagerFactory(new \Doctrine\DBAL\Schema\LegacySchemaManagerFactory());
        // DbalTracingPass:
        $a->setMiddlewares([0 => new \Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriverMiddleware(new \Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriverConnectionFactory(($this->privates['Sentry\\State\\HubInterface'] ?? $this->getHubInterfaceService())))]);
        // MiddlewarePass:
        $a->setMiddlewares([0 => $b]);

[...]
    }
```

It seems to me that the proper way to go here if possible, is to cooperate with the MiddlewarePass instead of fighting it, e.g. attempting to get TracingDriverMiddleware tagged properly.
Usually, the appropriate tag needed for MiddlewarePass to find middlewares should be added automatically through autowiring to anything implementing its `Middleware` interface, or marked through the `AsMiddleware` PHP attribute. The tagging is done by `DoctrineExtension`.

I don't know enough about autowiring and Symfony's DI system to answer why this isn't working:
- TracingDriverMiddleware (indirectly) implements DBAL v3's Middleware interface - maybe it would need to implement it directly?
- Manually adding the AsMiddleware attribute did not help, so maybe autowiring just ignores this? Or it's not configured correctly to look at it..

ste93cry:
Fixes #805. The problem that was reported is that the `MiddlewarePass` compiler pass in `DoctrineBundle` [adds](https://github.com/doctrine/DoctrineBundle/blob/fb22c9320383caa2829c32dc3b84c1b5e0f1ceaf/DependencyInjection/Compiler/MiddlewaresPass.php#L81-L83) a new call to `setMiddlewares()`, which means that all the middlewares added before get lost. Technically speaking, if you load `SentryBundle` after `DoctrineBundle`, it works as expected, because our code appends our middleware to the existing list instead of replacing it entirely. However, as noted by @fridtjof, a better integration approach is to tag our middleware as `doctrine.middleware` and let the Doctrine bundle do all the rest of the work to properly configure the connection.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
     },
     "require-dev": {
         "doctrine/dbal": "^2.13||^3.0",
-        "doctrine/doctrine-bundle": "^1.12||^2.5",
+        "doctrine/doctrine-bundle": "^2.6",
         "friendsofphp/php-cs-fixer": "^2.19||^3.40",
         "masterminds/html5": "^2.8",
         "phpstan/extension-installer": "^1.0",
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,11 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Parameter \\#1 \\.\\.\\.\\$arrays of function array_merge expects array, mixed given\\.$#"
-			count: 1
-			path: src/DependencyInjection/Compiler/DbalTracingPass.php
-
-		-
 			message: "#^Call to an undefined method Symfony\\\\Component\\\\Config\\\\Definition\\\\Builder\\\\TreeBuilder\\:\\:root\\(\\)\\.$#"
 			count: 1
 			path: src/DependencyInjection/Configuration.php
--- a/src/DependencyInjection/Compiler/DbalTracingPass.php
+++ b/src/DependencyInjection/Compiler/DbalTracingPass.php
@@ -9,7 +9,6 @@ use Sentry\SentryBundle\Tracing\Doctrine\DBAL\ConnectionConfigurator;
 use Sentry\SentryBundle\Tracing\Doctrine\DBAL\TracingDriverMiddleware;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
 
 final class DbalTracingPass implements CompilerPassInterface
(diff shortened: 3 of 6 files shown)


fridtjof:
Do note that this raises the required doctrine-bundle version to 2.6 (as opposed to the current ^2.5 listed in require-dev), as their tagged middleware support was only introduced then: https://github.com/doctrine/DoctrineBundle/releases/tag/2.6.0, https://github.com/doctrine/DoctrineBundle/pull/1472

Jean85 (approved):
This is a very interesting approach, and also forward compatible, thank you!

PS: do we have any E2E test to check if all of this is working?

ste93cry:
>Do note that this raises the required doctrine-bundle version to 2.6

Nice catch. Unfortunately, I realized that these changes are not compatible with version `1.12` of the bundle, which we still support. It seems that we declare the compatibility with that version mainly because of Symfony `3.4`, but since we nowadays require at least version `4.4` of the framework, I think we can just drop the support for the the older `1.x`...

>do we have any E2E test to check if all of this is working?

No, we don't. There are some E2E tests for other stuffs, but nothing specifically tailored at asserting that the tracing integrations work as expected. It's something we could definitely improve in the future though...

Issue #834, opened 2024-04-16
PHP Fatal error:  Declaration of Sentry\SentryBundle\Tracing\HttpClient\TraceableResponseForV4::getInfo(?string $type = null) must be compatible with Symfony\Contracts\HttpClient\ResponseInterface::getInfo(?string $type = null):


jeslopcru:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

5.0.0

### Steps to reproduce

I tried to Bump sentry/sentry-symfony from 4.14.0 to 5.0.0.
When I run PHP-cs-fixer or PHPStan or any other command 
Then the result is a failure
```
PHP Fatal error:  Declaration of Sentry\SentryBundle\Tracing\HttpClient\TraceableResponseForV4::getInfo(?string $type = null) must be compatible with Symfony\Contracts\HttpClient\ResponseInterface::getInfo(?string $type = null): mixed in /home/runner/work/phone-app/phone-app/vendor/sentry/sentry-symfony/src/Tracing/HttpClient/TraceableResponseForV4.php on line 15
Fatal error: Declaration of Sentry\SentryBundle\Tracing\HttpClient\TraceableResponseForV4::getInfo(?string $type = null) must be compatible with Symfony\Contracts\HttpClient\ResponseInterface::getInfo(?string $type = null): mixed in /home/runner/work/phone-app/phone-app/vendor/sentry/sentry-symfony/src/Tracing/HttpClient/TraceableResponseForV4.php on line 15
time=2024-04-15T14:04:36.151Z level=INFO msg="[phpstan] reported: https://github.com/DocPlanner/phone-app/runs/23830731474 (conclusion=success)"
```
I'm using `symfony/http-client-contracts` version 3.4.2 in composer.lock

### Expected result

A ok result when executing the command

This is signature of the method `getInfo`
https://github.com/symfony/contracts/blob/main/HttpClient/ResponseInterface.php#L108
```
public function getInfo(?string $type = null): mixed;
```

### Actual result

```
PHP Fatal error:  Declaration of Sentry\SentryBundle\Tracing\HttpClient\TraceableResponseForV4::getInfo(?string $type = null) must be compatible with Symfony\Contracts\HttpClient\ResponseInterface::getInfo(?string $type = null): mixed in /home/runner/work/phone-app/phone-app/vendor/sentry/sentry-symfony/src/Tracing/HttpClient/TraceableResponseForV4.php on line 15
[...]

PR #835, opened 2024-04-16
fix TraceableResponseForV4::getInfo  to be compatible with symfony-contracts in symfony-sentry 5.0.0

jeslopcru:
Hi!
I found this issue https://github.com/getsentry/sentry-symfony/issues/834 about a wrong signature in method `getInfo`
on `Sentry\SentryBundle\Tracing\HttpClient\TraceableResponseForV4` using version 5.0.0.

Here you can see the change to make compatible the method.

I hope that it will help, let me know if you need any update

### DIFF
--- a/src/Tracing/HttpClient/TraceableResponseForV4.php
+++ b/src/Tracing/HttpClient/TraceableResponseForV4.php
@@ -12,7 +12,7 @@ final class TraceableResponseForV4 extends AbstractTraceableResponse
     /**
      * {@inheritdoc}
      */
-    public function getInfo(string $type = null)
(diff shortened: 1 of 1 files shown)


Jean85 (changes requested):
I approved the CI just to show how this change is not feasible: this bundle supports back to PHP 7.2, so `mixed` is not allowed.

Thanks to PHP 7.2 having partial type widening though, this is perfectly valid code.

Jean85:
Thanks for the contribution anyway.

jeslopcru:
Thanks for the explanation! I will figure out how to fix my code to avoid this issue

PR #849, opened 2024-06-04
Pin PHPStan

Jean85:
This should fix the CI now and in the future, see #848 as an example.


stayallive (approved):

stayallive:
Thanks! Good call, phpstan is great but having it "auto" update is not a great situation.

PR #861, opened 2024-07-12
Add HandlerMessageBody to SentryException

### LINKED ISSUE #703: Add message id or content for exception thown in SF Messenger handlers

y4roc:
This PR should resolve #703.


Jean85 (changes requested):
This is not a feasible approach: throwing an exception is a no-go, also you're giving for granted that the message is in JSON (which is not even the default for Messenger).

y4roc:
`getMessage()` will always return an object. So normally, no exception will be thrown. I would add an option to deactivate the messageBody, because sometimes it will include secret datas.

In future it would be great to have a service, wich can hide some properties.

PR #864, opened 2024-07-23
ci: dependency review action

mdtro:
Add dependency review. Any new dependencies added with known high or critical severity will fail CI.


cleptric (approved):

Jean85:
How does this work, since there is no `composer.lock` committed?

PR #891, opened 2024-11-04
Set status for console command transactions

rodion-k:
Currently console command transactions are added with status unknown. As a result, the error rate in queues is incorrect if a message was produced in a console command. This change sets the status ok if the exit code is 0 and the status internal error otherwise.


cleptric:
Thanks!

PR #894, opened 2024-11-14
fix: Implicitly marking parameter as nullable is deprecated

NikoGrano:
Fixed PHP 8.4: Implicitly nullable parameter declarations deprecated.

[Ref](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated)


Jean85 (approved):

ruudk:
Thanks for this. Could this please be tagged? @Jean85

cleptric:
This PR didn't fix all incompatibilities, a followup was created as https://github.com/getsentry/sentry-symfony/pull/893. I'm currently looking into the last remaining CI failure, we'll get a release out before Thursday.

PR #912, opened 2025-02-26
Allow to install the bundle without Symfony security

Nyholm:
I have a semi small micro service running Symfony. We dont use any security feature (it is a private network). I would love to use this package but I also dont want to install dependencies not needed. 

I saw that the `symfony/security-core` and `symfony/security-http` packages are only used in `LoginListener`. I added check to remove that service if `symfony/security-core` is not installed. I also improved the logic between `src/Resources/config/services.xml` and the compiler pass. 

Merging this will not affect existing users, since they either have `symfony/security-core` required or they are not using it. 

There is one edge case though. Someone installed this package, configure security (without `symfony/security-bundle`) AND did not specify `symfony/security-core` in their composer.json. I find that extremely unlikely but I thought I would mention it.

### DIFF
--- a/src/DependencyInjection/Compiler/AddLoginListenerTagPass.php
+++ b/src/DependencyInjection/Compiler/AddLoginListenerTagPass.php
@@ -17,9 +17,17 @@ final class AddLoginListenerTagPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container): void
     {
+        if (!$container->hasDefinition(LoginListener::class)) {
+            return;
+        }
         $listenerDefinition = $container->getDefinition(LoginListener::class);
 
-        if (!class_exists(LoginSuccessEvent::class)) {
+        if (class_exists(LoginSuccessEvent::class)) {
+            $listenerDefinition->addTag('kernel.event_listener', [
+                'event' => LoginSuccessEvent::class,
+                'method' => 'handleLoginSuccessEvent',
+            ]);
+        } elseif (class_exists(AuthenticationSuccessEvent::class)) {
             $listenerDefinition->addTag('kernel.event_listener', [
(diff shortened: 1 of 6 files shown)


deluxetom (approved):

src/DependencyInjection/Compiler/AddLoginListenerTagPass.php
```diff
@@ -17,9 +17,17 @@ final class AddLoginListenerTagPass implements CompilerPassInterface
...
+            return;
+        }
         $listenerDefinition = $container->getDefinition(LoginListener::class);
 
-        if (!class_exists(LoginSuccessEvent::class)) {
+        if (class_exists(LoginSuccessEvent::class)) {
+            $listenerDefinition->addTag('kernel.event_listener', [
+                'event' => LoginSuccessEvent::class,
+                'method' => 'handleLoginSuccessEvent',
+            ]);
```
Jean85, line 29: Codecov is warning that this branch is not covered... is this still relevant? Or has it fallen outside of our supported Symfony versions?

Nyholm: Oh, you are correct. 

I messed up with `class_exists()` vs `interface_exists()`. 

I added a few more tests to cover this.

welcoMattic (approved):

cleptric (approved):
Thank you!

Nyholm:
Wohoo. Thank you for merging

cleptric:
Was part of [5.2.0](https://github.com/getsentry/sentry-symfony/releases/tag/5.2.0), released yesterday.

Nyholm:
Wohoo!

PR #948, opened 2025-09-09
fix: support named arguments in BufferFlushPass

kk-daniel:
This PR fixes "Undefined array key 0" on BufferFlushPass when using named arguments


Litarnus (approved):
Nice! Thanks for your contribution

PR #959, opened 2025-09-18
fix(logs): Allow passing string and enum Monolog level

HypeMC:
### Description

Currently, the constructor only accepts integers as the log level, e.g. by using the `Monolog\Logger` constants:

```yaml
services:
  Sentry\SentryBundle\Monolog\LogsHandler:
    arguments:
      - !php/const Monolog\Logger::INFO
```

This, however, isn't very flexible, especially since the Monolog constants are deprecated.
This PR expands the accepted types to also allow strings or the Monolog `Level` enum, like other handlers:

```yaml
services:
  Sentry\SentryBundle\Monolog\LogsHandler:
    arguments:
      - 'info'
```

```yaml
# or using PSR constants
services:
  Sentry\SentryBundle\Monolog\LogsHandler:
    arguments:
      - !php/const Psr\Log\LogLevel::INFO
```

```php
// or via PHP config
$container->services()
    ->set(\Sentry\SentryBundle\Monolog\LogsHandler::class)
    ->args([\Monolog\Level::Info]);
```

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -18,6 +18,20 @@
       <code><![CDATA[isMasterRequest]]></code>
     </UndefinedMethod>
   </file>
+  <file src="src/Monolog/LogsHandler.php">
+    <InvalidParamDefault>
+      <code><![CDATA[value-of<MonologLevel::VALUES>|value-of<MonologLevel::NAMES>|MonologLevel|PsrLogLevel::*]]></code>
+    </InvalidParamDefault>
+    <UndefinedClass>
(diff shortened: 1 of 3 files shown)


psalm-baseline.xml
```diff
@@ -18,6 +18,20 @@
       <code><![CDATA[isMasterRequest]]></code>
     </UndefinedMethod>
   </file>
+  <file src="src/Monolog/LogsHandler.php">
```
HypeMC, line 21: Seems to be a bug in Psalm when using const on enums: https://psalm.dev/r/a6f9298b82

PR #968, opened 2025-10-31
feat: add support for symfony 8

Litarnus:
We need to cleanup `composer.json` and the CI config once Symfony 8 is properly released to use the stable version instead of `BETA` before merging this.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,8 +43,6 @@ jobs:
             symfony-version: 6.*
           - php: '7.4'
             symfony-version: 6.*
-          - php: '8.0'
-            symfony-version: 6.*
           - php: '7.2'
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -54,6 +55,20 @@ jobs:
             symfony-version: 7.*
           - php: '8.4'
             symfony-version: 4.4.*
+          - php: '7.2'
+            symfony-version: 8.0
```
cleptric:
```suggestion
            symfony-version: 8.*
```

etc.

Schneider95:
Hi,

any idea if the support for Symfony 8 will be added to the 4.x branch ? 

Thanks in advance.

stayallive (approved):

PR #1043, opened 2026-07-12
fix: start runtime context before the router and security firewall

### LINKED ISSUE #1042: Pre-context log records leak across requests on persistent workers (RuntimeContextListener starts after router/firewall)
### Environment

- sentry/sentry `4.29.0`, sentry/sentry-symfony `5.10.0`
- Symfony 8.0, PHP 8.5
- FrankenPHP worker mode (persistent worker; same applies to RoadRunner)

### Summary

The runtime-context isolation introduced in #997 starts the per-request context in `RuntimeContextListener` at `kernel.request` **priority 6** — which is *after* Symfony's `RouterListener` (priority **32**) and the security firewall (priority **8**).

Anything written to the current hub before priority 6 lands on the shared **base hub**, and `RuntimeContextManager::createHubFromBaseHub()` clones the base hub's scope into every new context. On a persistent worker the base hub lives for the whole process, so this state **accumulates and leaks into every subsequent request's events**.

Two concrete leak vectors, both with vanilla wiring:

1. **Breadcrumbs** (with the documented `Sentry\Monolog\BreadcrumbHandler` monolog setup): `RouterListener` logs `Matched route "..."` and authenticators log e.g. `Authenticator failed` during the firewall — both before priority 6. These become breadcrumbs on the base hub and then appear on the events of *unrelated later requests*, up to the 100-breadcrumb cap.
2. **User misattribution** (with `send_default_pii: true`): `LoginListener::handleLoginSuccessEvent()` fires during the firewall (before priority 6) and writes the authenticated user to the base hub. `handleKernelRequestEvent` (priority 0) re-applies the token user inside the context on later authenticated requests, but **anonymous requests on the same worker inherit the baseline user from the clone** — errors get attributed to whoever logged in last on that worker.

### Reproduction (SDK only, mirrors the listener ordering)

```php
\Sentry\init(['default_integrations' => false]);

// REQUEST 1: router/firewall log records become breadcrumbs BEFORE startContext() (priority 6)
SentrySdk::getCurrentHub()->addBreadcrumb(new Breadcrumb(
    Breadcrumb::LEVEL_INFO, Breadcrumb::TYPE_DEFAULT, 'security', 'REQUEST-1: Authenticator failed'
));
SentrySdk::startContext();   // kernel.request, priority 6
SentrySdk::endContext();     // kernel.terminate, priority -128

// REQUEST 2, same worker process:
SentrySdk::getCurrentHub()->addBreadcrumb(new Breadcrumb(
    Breadcrumb::LEVEL_INFO, Breadcrumb::TYPE_DEFAULT, 'security', 'REQUEST-2: Authenticator failed'
));
SentrySdk::startContext();

$event = null;
SentrySdk::getCurrentHub()->configureScope(function (Scope $scope) use (&$event): void {
    $event = $scope->applyToEvent(Event::createEvent());
});

foreach ($event->getBreadcrumbs() as $breadcrumb) {
    echo $breadcrumb->getMessage(), PHP_EOL;
}
```

Output:

```
REQUEST-1: Authenticator failed   <-- leaked from the previous request
REQUEST-2: Authenticator failed
```

### Expected behavior

Each request's events only carry state from that request — the point of #997.

### Suggested fix

Start the context before the router and the firewall, e.g. `kernel.request` priority `512` (any value above 32 fixes the vanilla vectors; 512 also clears `ValidateRequestListener` at 256 and `SessionListener` at 128). Nested `startContext()` calls are already a documented no-op in `RuntimeContextManager`, so the change is minimal. PR incoming.

**Workaround** for affected users until then: register an app listener on `kernel.request` at a high priority that calls `SentrySdk::startContext()` — the bundle's own later start becomes a no-op.

JanMikes:
Fixes #1042

`RuntimeContextListener` started the per-request runtime context at `kernel.request` priority **6** — after `RouterListener` (**32**) and the security firewall (**8**). Records written to the hub before the context starts land on the shared base hub, whose scope `RuntimeContextManager::createHubFromBaseHub()` clones into every new context. On persistent workers (FrankenPHP, RoadRunner) that base hub lives for the whole process, so pre-context breadcrumbs (`Matched route "..."`, authenticator log records via the documented `BreadcrumbHandler` setup) and the `LoginSuccessEvent` user write accumulated there and leaked into every subsequent request's events. See #1042 for a minimal SDK-only reproduction.

### Change

- Start the context at `kernel.request` priority **512**, before the router and the firewall, so all request-phase records land in the per-request context. Nested `startContext()` calls are a documented no-op in `RuntimeContextManager`, so no other behavior changes; the terminate/reset lifecycle is untouched.
- Why 512: any value above 32 fixes the vanilla leak vectors; 512 also runs ahead of `ValidateRequestListener` (256) and `SessionListener` (128) while leaving headroom for apps that deliberately want a listener to run pre-context. Happy to adjust if you prefer a different value.
- Updated the DI test expectation and added a CHANGELOG entry.

All `tests/DependencyInjection` + `RuntimeContextListenerTest` tests pass (171 tests, 358 assertions).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Unreleased
+
+### Bug fixes
+
+- Start the runtime context before the router and the security firewall (`kernel.request` priority `512` instead of `6`). Log records emitted by those listeners (e.g. "Matched route", authenticator messages) become breadcrumbs on the shared base hub before the context starts; since every runtime context clones the base hub's scope, they leaked across requests on persistent workers (FrankenPHP, RoadRunner). [(#1043)](https://github.com/getsentry/sentry-symfony/pull/1043)
+
 ## 5.10.0
 
 The Sentry SDK team is happy to announce the immediate availability of Sentry Symfony SDK v5.10.0.
--- a/src/Resources/config/services.yaml
+++ b/src/Resources/config/services.yaml
@@ -15,11 +15,7 @@ services:
   Sentry\SentryBundle\EventListener\RuntimeContextListener:
     arguments: ['@Sentry\State\HubInterface']
     tags:
-      # The runtime context must start before any listener that can write to the hub:
-      # the router (priority 32) and the security firewall (priority 8) emit log records
-      # that monolog integrations turn into breadcrumbs. Before the context starts these
-      # land on the shared base hub, whose scope is cloned into every new context - on
-      # persistent workers (FrankenPHP, RoadRunner) they would leak across requests.
+      # Use a very high priority to start a context before anything else can emit logs
       - { name: kernel.event_listener, event: kernel.request, method: handleKernelRequestEvent, priority: 512 }
       - { name: kernel.event_listener, event: kernel.terminate, method: handleKernelTerminateEvent, priority: -128 }
       - { name: kernel.reset, method: reset }
--- a/src/Resources/config/services.yaml
+++ b/src/Resources/config/services.yaml
@@ -15,7 +15,12 @@ services:
   Sentry\SentryBundle\EventListener\RuntimeContextListener:
     arguments: ['@Sentry\State\HubInterface']
     tags:
-      - { name: kernel.event_listener, event: kernel.request, method: handleKernelRequestEvent, priority: 6 }
+      # The runtime context must start before any listener that can write to the hub:
+      # the router (priority 32) and the security firewall (priority 8) emit log records
+      # that monolog integrations turn into breadcrumbs. Before the context starts these
+      # land on the shared base hub, whose scope is cloned into every new context - on
+      # persistent workers (FrankenPHP, RoadRunner) they would leak across requests.
+      - { name: kernel.event_listener, event: kernel.request, method: handleKernelRequestEvent, priority: 512 }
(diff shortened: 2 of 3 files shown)


Litarnus (commented):
Please remove the Changelog section, we usually add those things before releasing.

Also the comment in `services.yaml` can be much shorter I think. Something like 
> Use a very high priority to start a context before anything else can emit logs

should make this sufficiently clear.

Rest looks good, thank you!

Litarnus:
Oh and if you merge master the failing tests will pass again

JanMikes:
Thanks for the review @Litarnus! All addressed:

- Changelog entry removed (dropped while merging master, which is now merged in — 5.11.0)
- services.yaml comment shortened to your suggested wording
- Added `testRuntimeContextListenerStartsBeforeAnyListenerThatCanEmitLogs` asserting the request priority stays above `RouterListener`'s actual subscribed priority, so the invariant is captured beyond the pinned tag value — happy to drop it if you prefer the minimal diff

Litarnus (approved):
Looks good! Thanks for the contribution

